Issue #836, opened 2026-04-20
Issues with deploying libraries to integrator cloud


madushajg:
### Description

Following issues were identified when deploying libraries to integrator cloud.
1. Libraries are not automatically selected in deploy view
2. `component/create` failed with `Request failed: Component type is not supported` when trying to deplo a library
3. The notifications and overview incorrectly use the term 'integration' for libraries

### Steps to Reproduce

1. Create a Project
2. Add a library and an integration
3. Click on `Deploy` 

### Version

v5.0.0-alpha6

PR #862, opened 2026-04-20
Auto-select library components in Deploy View

madushajg:
## Purpose
$title
Addresses: https://github.com/wso2/product-integrator/issues/836


madushajg:
@coderabbitai review

kaje94:
Shouldn't we simply hide the library components from the deploy view (since we are not actually deploying them)?

madushajg:
> Shouldn't we simply hide the library components from the deploy view (since we are not actually deploying them)?

Even though we don’t deploy library components, users still need to push them to GitHub because integrations depend on those libraries.

Also with the cloud support for library components, users may eventually be able to view libraries within the project in Cloud, similar to how integrations are shown.

Because of this, hiding library components from the deploy view might create confusion, since they are still part of the project lifecycle and dependency structure.

kaje94:
> Even though we don’t deploy library components, users still need to push them to GitHub because integrations depend on those libraries.

For this, we simply need to show a warning message if dependency library directories are not pushed to Git

> Also with the cloud support for library components, users may eventually be able to view libraries within the project in Cloud, similar to how integrations are shown.
> 
> Because of this, hiding library components from the deploy view might create confusion, since they are still part of the project lifecycle and dependency structure.

IMO, showing libraries at the same level as integrations will create more confusion.

Lets have an offline chat to discuss this

kaje94 (approved):

PR #1220, opened 2026-04-26
Use app product version for update check current version

chiranSachintha:
## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/wi/wi-extension/src/services/productUpdateServiceClient.ts
+++ b/wi/wi-extension/src/services/productUpdateServiceClient.ts
@@ -39,6 +39,7 @@ const DEFAULT_REQUEST_TIMEOUT_MS = 10000;
 const LAST_CHECKED_AT_KEY = "wso2-integrator.updates.lastCheckedAt";
 const LAST_NOTIFIED_VERSION_KEY = "wso2-integrator.updates.lastNotifiedVersion";
 const LAST_NOTIFIED_INSTALLED_VERSION_KEY = "wso2-integrator.updates.lastNotifiedInstalledVersion";
+const APP_VERSION_ENV_KEY = "WSO2_INTEGRATOR_VERSION";
 
 interface ProductJsonShape {
 	wiversion?: string;
@@ -153,6 +154,11 @@ export class ProductUpdateServiceClient {
 	}
 
 	private getInstalledVersion(): string | undefined {
+		const appVersionFromEnv = this.getAppVersionFromEnv();
+		if (appVersionFromEnv) {
+			return appVersionFromEnv;
+		}
+
 		const appVersion = this.getAppVersionFromProductJson();
 		if (appVersion) {
 			return appVersion;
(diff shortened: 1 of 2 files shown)


wi/wi-extension/src/services/productUpdateServiceClient.ts
```diff
@@ -157,6 +170,22 @@ export class ProductUpdateServiceClient {
 		return extension?.packageJSON?.version;
 	}
 
+	private getAppVersionFromProductJson(): string | undefined {
```
gigara: Can't we use `vscode.version`?

chiranSachintha: Reason to use `wiversion` is we set the version from `integrator_version` in ci/build/update-product.sh.
vscode.version is VS Code engine version, not WSO2 Integrator app version. RIght?
https://github.com/wso2/product-integrator/blob/8108a62df1f940edeac66c8ee4208f3dde8e5cb1/ci/build/update-product.sh#L64

keizer619: As this is the version definition lets use this reference 
https://github.com/wso2/product-integrator/blob/5.0.x/ci/build/component-versions.properties#L1

chiranSachintha: IMO `component-versions.properties` is used only to populate `wiversion `during build. In runtime safe source is `wiversion`. @kaumini @gigara WDYT?

wi/wi-extension/src/services/productUpdateServiceClient.ts
```diff
@@ -157,6 +169,22 @@ export class ProductUpdateServiceClient {
...
+		try {
+			const productJsonPath = path.join(vscode.env.appRoot, "product.json");
```
gigara: This logic will not work in vscode environment. A better option is to run this update checker only in the WSO2 Integrator environment. Also, we can pass the Integrator version as an env and read it here to avoid file read.

chiranSachintha: I have updated the logic. please check

wi/wi-extension/src/services/productUpdateServiceClient.ts
```diff
@@ -147,6 +154,17 @@ export class ProductUpdateServiceClient {
 	}
 
 	private getInstalledVersion(): string | undefined {
+		const appVersionFromEnv = this.getAppVersionFromEnv();
+		if (appVersionFromEnv) {
+			return appVersionFromEnv;
+		}
+
+		const appVersion = this.getAppVersionFromProductJson();
```
gigara: Why do we need fallbacks? We need the update checker only in the Integrator App, right? We don't need that if the extension is installed in VS Code.

chiranSachintha: Updated.

gigara (approved):

PR #1953, opened 2026-07-30
fix(wi): authenticate the webview websocket bridge and allowlist runCommand

gigara:
## Purpose

The embedded welcome page starts the giga-bridge websocket backend on activation (`BridgeLayer.startWebSocketServer`), but the transport manager was created **without an auth token**, and the `runCommand` route forwarded arbitrary command ids straight to `commands.executeCommand`.

Any client that could reach the port was therefore able to drive the extension — execute commands, read cloud credentials, clone repositories, write project files — and additionally received every broadcast event (`STATE_CHANGED`, `AUTH_STATE_CHANGED`, download/clone progress) simply by connecting and staying quiet.

The listening socket also bound to all interfaces, which is fixed in the companion library change (wso2/vscode-extensions#2435) that this PR picks up.

## Goals

1. Make reaching the port insufficient — connections must prove they are the intended webview.
2. Stop `runCommand` from being a general-purpose command-execution primitive, even for a caller that *is* authenticated.

## Approach

**1. Token-gated websocket (`BridgeLayer.ts`)** — the transport manager is created with a per-channel `authToken` (32 random bytes). The bridge validates it during the websocket handshake and rejects non-matching upgrades, so unauthenticated clients never connect and therefore cannot issue requests *or* observe broadcasts.

Proxy mode is deliberately left ungated: it runs in-process over the trusted webview `postMessage` channel and is not exposed on a port.

**2. Token delivery (`WITransportBootstrap`, `WsClient.ts`)** — the token travels to the webview as `wsToken` and is passed to the transport adapter, which sends it during the handshake. No new extension plumbing was required — `webviewManager` already serializes the whole bootstrap object into the page. The query-parameter form is kept only for browser-based dev runs.

**3. `runCommand` allowlist (`ws-managers/main/ws-manager.ts`)** — command ids outside `ALLOWED_WEBVIEW_COMMANDS` are refused and logged. The list was derived from the actual `wsClient.runCommand` call sites in `wi-webviews` (13 of them), not guessed:

| Source | Commands |
|---|---|
| `WICommandIds` | `SignIn`, `SignOut`, `CancelSignIn`, `CloneProject` |
| Runtime setup | `ballerina.setup-ballerina` |
| Workbench | `workbench.action.openRecent`, `workbench.action.reloadWindow`, `workbench.scm.focus`, `git.push` |

This layer matters because it still holds if the webview itself is compromised (e.g. XSS in welcome content) — a transport token cannot defend against that.

**Note on design:** authentication is enforced at the **handshake**, not per request. That is why no token field was added to `WIBridgeRequest` — rejected clients never open a connection, which also closes the broadcast-observation path that a per-request check would leave open.

## Submodule bump

Includes `external/wso2-vscode-extensions`: `fae80157d` → `be1a5b98a`. This is **required**, not incidental — the `authToken` / `token` / `wsToken` options do not exist at the previous pin, so the code cannot compile without it. That commit is the merge of wso2/vscode-extensions#2435 (loopback binding + handshake token auth in `@wso2/webview-giga-bridge`).

## User stories

As a developer running WSO2 Integrator, another user on my network (or a web page I merely visit) must not be able to connect to the extension's local bridge and execute commands or read my cloud credentials.

## Release note

Fixed an unauthenticated local websocket bridge in the WSO2 Integrator welcome page and restricted the commands the webview may invoke.

## Documentation

N/A — no user-facing behaviour or configuration changes. The token is generated and consumed internally.

## Training

N/A

## Certification

N/A — internal transport hardening, no impact on certification exam content.

## Marketing

N/A

## Automation tests

- Unit tests
  > No unit-test harness exists for `BridgeLayer` / `MainWsManager` in this package, so none were added. Verification was done by build and typecheck (below).
- Integration tests
  > None added. The websocket path is exercised only in `WEB_VIEW_DEV_MODE` and by the embedded welcome flow, neither of which currently has automated coverage.

**Verification performed:**
- `wi-extension` typechecks clean (0 errors)
- `wi-core` builds and emits `wsToken?: string`
- `wi-webviews` reports 0 errors in `WsClient.ts`
- `WICommandIds` confirmed to export exactly the four ids referenced by the allowlist

**Known gap:** `wi-webviews` still reports 34 errors in this local partial build — 26 `Cannot find module '@wso2/ui-toolkit'` plus 8 cascading implicit-any in two of its consumers. These are pre-existing and unrelated: none are in files touched here, both implicit-any files import `ui-toolkit`, and `ui-toolkit` was not built locally. A full `rush build` (as CI runs) builds it first.

## Security checks

- Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? **yes**
- Ran FindSecurityBugs plugin and verified report? **N/A** — TypeScript codebase, not Java
- Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? **yes** — the token is generated at runtime via `randomBytes(32)` and never persisted or committed

## Samples

N/A

## Related PRs

- wso2/vscode-extensions#2435 — the `@wso2/webview-giga-bridge` change this depends on (loopback bind + handshake token auth). Already merged; picked up here via the submodule bump.

## Migrations (if applicable)

None. No persisted state or configuration is affected.

## Test environment

macOS (darwin 25.5.0), Node 22.21.1, Rush 5.177.1 / pnpm 10.11.0.

## Learning

The non-obvious constraint was that `brace-expansion`-style "just take the latest fixed version" reasoning does not apply to transport auth: enforcing the token per **request** (as the Ballerina `DefaultServer` does) still lets an unauthenticated client connect and passively receive broadcast events. Gating at the handshake closes both paths at once, which is why the library change exposes `verifyClient` rather than a per-message check.

## Summary by CodeRabbit

* **Security**
  * Added token-based authentication for embedded webview WebSocket connections.
  * Restricted webview-invoked actions to approved commands.
  * Strengthened embedded Welcome content with per-render CSP nonces.

* **New Features**
  * Added an improved empty-workspace Welcome experience with WSO2 login support.

* **Reliability**
  * Ensured webviews wait for bridge startup and report startup errors.

* **Compatibility**
  * Updated bundled WSO2 VS Code extensions.

### DIFF
--- a/patches/gettingStarted.diff
+++ b/patches/gettingStarted.diff
@@ -247,7 +247,7 @@ Index: product-integrator/lib/vscode/src/vs/workbench/browser/parts/editor/wso2E
 ===================================================================
 --- /dev/null
 +++ product-integrator/lib/vscode/src/vs/workbench/browser/parts/editor/wso2EditorGroupWatermark.ts
-@@ -0,0 +1,667 @@
+@@ -0,0 +1,691 @@
 +/*---------------------------------------------------------------------------------------------
 + *  Copyright (c) Microsoft Corporation. All rights reserved.
 + *  Licensed under the MIT License. See License.txt in the project root for license information.
@@ -264,7 +264,8 @@ Index: product-integrator/lib/vscode/src/vs/workbench/browser/parts/editor/wso2E
 +// eslint-disable-next-line local/code-import-patterns
 +import { IWebviewElement, IWebviewService } from '../../../contrib/webview/browser/webview.js';
 +// eslint-disable-next-line local/code-import-patterns
-+import { asWebviewUri } from '../../../contrib/webview/common/webview.js';
++import { asWebviewUri, webviewRootResourceAuthority } from '../../../contrib/webview/common/webview.js';
++import { generateUuid } from '../../../../base/common/uuid.js';
 +// eslint-disable-next-line local/code-import-patterns
 +import { WebviewThemeDataProvider } from '../../../contrib/webview/browser/themeing.js';
 +import { IExtensionService } from '../../../services/extensions/common/extensions.js';
@@ -275,6 +276,12 @@ Index: product-integrator/lib/vscode/src/vs/workbench/browser/parts/editor/wso2E
 +	mode: 'proxy' | 'websocket';
 +	wsServer: string;
 +	wsPort: number;
++	/**
++	 * Token the webview presents on the bridge websocket handshake. The extension
++	 * refuses upgrades without it, so this must survive into the injected
++	 * bootstrap or the embedded view cannot connect.
++	 */
++	wsToken?: string;
 +}
 +
 +interface IEmbeddedWelcomeThemeData {
@@ -612,10 +619,27 @@ Index: product-integrator/lib/vscode/src/vs/workbench/browser/parts/editor/wso2E
 +		const serializedBootstrap = JSON.stringify(bootstrap).replace(/</g, '\\u003c');
 +		const serializedThemeData = JSON.stringify(themeData).replace(/</g, '\\u003c');
 +
++		// Mirrors the policy the WSO2 Integrator extension applies to this same
++		// bundle when it hosts it in an extension webview. Inline scripts are
++		// authorized by a per-render nonce; `connect-src` allows the loopback
++		// bridge socket this view talks to, and `img-src https:` covers the
++		// remotely hosted sample icons the welcome page renders.
++		const nonce = generateUuid();
++		const cspSource = `https://*.${webviewRootResourceAuthority}`;
++		const csp = [
++			`default-src 'none'`,
++			`script-src 'nonce-${nonce}' ${cspSource}`,
++			`style-src ${cspSource} 'unsafe-inline'`,
++			`font-src ${cspSource} data:`,
++			`img-src ${cspSource} data: https:`,
++			`connect-src ${cspSource} ws://127.0.0.1:*`
++		].join('; ');
++
 +		return `<!DOCTYPE html>
 +				<html lang="en">
 +					<head>
 +						<meta charset="utf-8">
++						<meta http-equiv="Content-Security-Policy" content="${csp}">
 +						<meta name="viewport" content="width=device-width,initial-scale=1,shrink-to-fit=no">
 +					<title>WSO2 Integrator</title>
 +					<style>
@@ -741,7 +765,7 @@ Index: product-integrator/lib/vscode/src/vs/workbench/browser/parts/editor/wso2E
 +					</style>
 +				</head>
 +				<body>
-+					<script>
++					<script nonce="${nonce}">
 +						(function () {
 +							var themeData = ${serializedThemeData};
 +							var rootStyle = document.documentElement.style;
@@ -777,7 +801,7 @@ Index: product-integrator/lib/vscode/src/vs/workbench/browser/parts/editor/wso2E
 +						<div class="embedded-welcome-loader-text">Loading WSO2 Integrator...</div>
 +					</div>
 +					<div id="root"></div>
-+					<script>
++					<script nonce="${nonce}">
 +						window.__WI_BRIDGE_BOOTSTRAP = ${serializedBootstrap};
 +						window.__WI_EMBEDDED_WELCOME_HIDE_LOADER = function () {
 +							document.body.classList.add('embedded-welcome-ready');
@@ -796,7 +820,7 @@ Index: product-integrator/lib/vscode/src/vs/workbench/browser/parts/editor/wso2E
 +							}
 +						});
 +					</script>
-+					<script defer src=${JSON.stringify(scriptUri)}></script>
++					<script defer nonce="${nonce}" src=${JSON.stringify(scriptUri)}></script>
 +					</body>
 +				</html>`;
 +	}
--- a/wi/wi-extension/src/stateMachine.ts
+++ b/wi/wi-extension/src/stateMachine.ts
@@ -278,7 +278,7 @@ const stateMachine = createMachine<MachineContext>({
                 });
             }
             vscode.commands.executeCommand('setContext', 'WI.projectType', 'none');
-            void context.webviewManager.showWelcome();
+            void context.webviewManager.showWelcome().catch((error) => ext.logError("Failed to show welcome webview", error as Error));
             context.currentView = ViewType.WELCOME;
         },
         registerConfigChangeListener: (context) => {
@@ -477,7 +477,7 @@ export const StateMachine = {
         // Get the webview manager from context and show the view
         const context = stateService.getSnapshot().context;
         if (context.webviewManager) {
-            void context.webviewManager.show(view);
+            void context.webviewManager.show(view).catch((error) => ext.logError("Failed to show webview", error as Error));
         } else {
             ext.log('WebviewManager not available in state machine context');
         }
--- a/wi/wi-extension/src/stateMachine.ts
+++ b/wi/wi-extension/src/stateMachine.ts
@@ -278,7 +278,7 @@ const stateMachine = createMachine<MachineContext>({
                 });
             }
             vscode.commands.executeCommand('setContext', 'WI.projectType', 'none');
-            context.webviewManager.showWelcome();
+            void context.webviewManager.showWelcome();
             context.currentView = ViewType.WELCOME;
         },
         registerConfigChangeListener: (context) => {
@@ -477,7 +477,7 @@ export const StateMachine = {
         // Get the webview manager from context and show the view
         const context = stateService.getSnapshot().context;
         if (context.webviewManager) {
-            context.webviewManager.show(view);
+            void context.webviewManager.show(view);
         } else {
             ext.log('WebviewManager not available in state machine context');
         }
--- a/wi/wi-extension/src/webviewManager.ts
+++ b/wi/wi-extension/src/webviewManager.ts
@@ -117,17 +117,21 @@ export class WebviewManager {
 
 	constructor(private projectUri: string) {
 		if (process.env.WEB_VIEW_DEV_MODE === "true") {
-			try {
-				const bootstrap = BridgeLayer.startWebSocketServer(projectUri);
-				vscode.window.showInformationMessage(
-					`Webview bridge server started on ws://${bootstrap.wsServer}:${bootstrap.wsPort}`,
-				);
-			} catch (error) {
-				ext.logError("Failed to start bridge server", error as Error);
-				vscode.window.showErrorMessage(
-					`Failed to start bridge server: ${error instanceof Error ? error.message : String(error)}`,
-				);
-			}
+			// The port is only known once the server is bound, and a constructor
+			// cannot await, so report the outcome when the start settles.
+			void BridgeLayer.startWebSocketServer(projectUri).then(
+				(bootstrap) => {
+					vscode.window.showInformationMessage(
+						`Webview bridge server started on ws://${bootstrap.wsServer}:${bootstrap.wsPort}`,
+					);
+				},
+				(error: unknown) => {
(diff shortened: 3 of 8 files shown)


gigara:
⚠️ **Do not merge before wso2/vscode-extensions#2442.**

While verifying the token gate empirically (by actually starting the server rather than only typechecking), I found that #2435 — the library change this PR depends on — has a regression: adding `host` for the loopback bind moved Node onto the `listen(port, host)` path, which resolves the host before binding, so `server.address()` is null in the same tick and the backend threw `Failed to allocate the WebSocket port.` on **every** start.

#2442 fixes it by reading the port from the `listening` event, which makes `startWebSocketServer()` and `switchMode('websocket')` return promises.

Pushed `ac9cc8b` to adapt:
- `BridgeLayer.startWebSocketServer` is now async and awaits the transport start before returning the bootstrap
- the embedded-welcome bootstrap command awaits it
- `WebviewManager`'s dev-mode start reports success/failure when the start settles, since a constructor cannot await and the call no longer throws synchronously

`wi-extension` typechecks clean (0 errors) against the fixed library.

**Two things still required before merge:**
1. The submodule pointer here is still `be1a5b98a`, which contains the regression. It must be moved to #2442's merge commit — I could not point it at the fix yet because that commit only exists on a fork branch and would not be fetchable from the submodule URL.
2. This has not been run in the app. The bug that started this was invisible to typechecks, so the welcome page should be opened once to confirm the webview actually connects.

gigara:
✅ **Dependency satisfied — this is no longer blocked.**

wso2/vscode-extensions#2442 merged as `29725ae50`, and `4611b5f` moves the submodule pin from `be1a5b98a` to it. The earlier pin bound the bridge to loopback but read the assigned port synchronously, throwing `Failed to allocate the WebSocket port.` on every start; #2442 reads it from the `listening` event.

Verified at the new pin: `wi-extension` typechecks clean (0 errors), including the awaited `startWebSocketServer()` and the `show()` change that waits for the bound port before rendering the webview HTML.

Still worth doing before merge: this has not been exercised in the running app. The two most serious bugs found while building this — the port regression and a `wsPort: 0` render race — were both invisible to typechecks and to CI, so opening the welcome page once is the check that actually confirms the webview connects.

patches/gettingStarted.diff
```diff
@@ -679,6 +686,23 @@ Index: product-integrator/lib/vscode/src/vs/workbench/browser/parts/editor/wso2E
...
++		const nonce = generateUuid();
++		const cspSource = `https://*.${webviewRootResourceAuthority}`;
++		const csp = [
++			`default-src 'none'`,
++			`script-src 'nonce-${nonce}' ${cspSource}`,
++			`style-src ${cspSource} 'unsafe-inline'`,
++			`font-src ${cspSource} data:`,
++			`img-src ${cspSource} data: https:`,
++			`connect-src ${cspSource} ws://127.0.0.1:*`
++		].join('; ');
```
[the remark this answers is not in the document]
gigara: Pinning to `bootstrap.wsPort` would break this page. The wildcard is load-bearing, not laziness.

This webview does not talk to only one bridge. It also hosts the **federated MI and BI forms**, added in #1952, and each of those connects to a *different* extension's websocket server on its own OS-allocated port — resolved at runtime, after this HTML and its CSP have already been generated:

```ts
// wi-webviews/src/views/creationView/federation/RemoteMiProjectForm.tsx:88
const coords = await wsClient.getMiFormWsBootstrap();
const bootstrap: MiBridgeBootstrap = {
    mode: "websocket",
    wsServer: coords.host,
    wsPort: coords.port,   // the MI extension's own server, not bootstrap.wsPort
    token: coords.token,
};
window.__MI_BRIDGE_BOOTSTRAP = bootstrap;
```

`bootstrap.wsPort` here is the **Integrator's** bridge only. The MI form's port comes from `getMiFormWsBootstrap`, and the BI form's from `ballerina.getBiFormWsBootstrap` on the Ballerina extension — both are `wsPort: 0` servers whose ports the OS assigns when those extensions start. There is no point during `getEmbeddedWelcomeHtml` at which they are knowable, and they change every session.

This is why the extension-side policy I mirrored carries the same wildcard, with the same reasoning recorded next to it:

```ts
// wi-extension/src/webviewManager.ts
// ws://127.0.0.1:* — the embedded BI form connects to the Ballerina
// extension's WS server (OS-allocated loopback port) for project RPCs.
`connect-src ${webview.cspSource} ws://127.0.0.1:*${devConnect}`,
```

Narrowing the app-side copy would make the two hosts diverge for identical content, and the embedded forms would fail only in the app — the harder of the two places to notice.

On `wsServer` being hardcoded: fair as written, but substituting it would not help. Every one of these servers binds loopback — `DEFAULT_WS_HOST = '127.0.0.1'` in the bridge library, which no consumer overrides — so `bootstrap.wsServer` is `127.0.0.1` in practice, and using it would still leave the MI/BI hosts uncovered since they are not in this bootstrap at all. If a server ever bound `::1`, the fix would be to allow that host explicitly rather than to derive one host from a bootstrap that does not describe the other two endpoints.

What the CSP does buy here, and what the token complements: `default-src 'none'` plus a nonce-gated `script-src` means nothing but our own bundles executes, and `connect-src` is still restricted to loopback — so the token embedded in this page cannot be shipped to a remote origin. Confining it further to one port would need the federated coordinates to be known up front, which would be a design change to how those forms are mounted.

wi/wi-extension/src/webviewManager.ts
```diff
@@ -117,17 +117,21 @@ export class WebviewManager {
...
+						`Webview bridge server started on ws://${bootstrap.wsServer}:${bootstrap.wsPort}`,
+					);
+				},
+				(error: unknown) => {
+					ext.logError("Failed to start bridge server", error as Error);
+					vscode.window.showErrorMessage(
+						`Failed to start bridge server: ${error instanceof Error ? error.message : String(error)}`,
+					);
+				},
+			);
```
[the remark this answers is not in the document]
gigara: Fixed in `bad2e79` (before the latest sync; flagging here since the thread is still open).

The finding was correct: the dev-mode bridge was started from the `WebviewManager` constructor, which cannot await, so `show()` could render the HTML — which embeds the bridge coordinates — before the port was bound. `WsClient` never refreshes its bootstrap, so the panel would have kept retrying `wsPort: 0` for its lifetime.

The startup is now kept on the instance and awaited before rendering:

```ts
/** Resolves once the dev-mode bridge server is bound. */
private bridgeReady: Promise<void> = Promise.resolve();

public async show(viewType: ViewType = ViewType.WELCOME): Promise<void> {
    // The webview HTML embeds the bridge coordinates, and the dev-mode server
    // is assigned its port asynchronously, so wait before rendering.
    await this.bridgeReady;
```

`bridgeReady` resolves on failure as well, so a bridge that cannot start still renders the panel and surfaces the error rather than leaving the view blank. `show()` and `showWelcome()` now return promises, and the two state-machine call sites are marked `void`, matching their existing fire-and-forget use.

`wi-extension` typechecks clean (0 errors), and CI on this PR is green.

wi/wi-extension/src/stateMachine.ts
```diff
@@ -278,7 +278,7 @@ const stateMachine = createMachine<MachineContext>({
                 });
             }
             vscode.commands.executeCommand('setContext', 'WI.projectType', 'none');
-            context.webviewManager.showWelcome();
+            void context.webviewManager.showWelcome();
```
NipunaRanasinghe: `show()` is now async, so `void` here means a rejection becomes an unhandled promise rejection with nothing logged. Previously a synchronous throw from `show()` (e.g. `createWebviewPanel`, or resource resolution in `getWebviewContent`) propagated into the xstate action; now it disappears silently.

Suggest attaching the repo's existing error sink:

```ts
void context.webviewManager.showWelcome().catch((error) => ext.logError("Failed to show welcome webview", error as Error));
```

Same applies to `void context.webviewManager.show(view)` at line 480.

wi/wi-extension/src/ws-managers/main/ws-manager.ts
```diff
@@ -140,6 +160,13 @@ export class MainWsManager implements WIVisualizerAPI {
     }
 
     async runCommand(props: RunCommandRequest): Promise<RunCommandResponse> {
+        if (!ALLOWED_WEBVIEW_COMMANDS.has(props.command)) {
+            // Forwarding arbitrary ids would make this a general command-execution
+            // primitive for anything that can reach the bridge.
+            ext.logError(`Blocked disallowed webview command: ${props.command}`);
+            return { success: false, error: `Command not allowed: ${props.command}` };
```
NipunaRanasinghe: Returning `{ success: false }` makes a blocked command a silent no-op in the UI. No `runCommand` call site in `wi-webviews` checks `success` — and it couldn't usefully, since the success path returns whatever `commands.executeCommand` returns (`undefined` for most of these), not a `RunCommandResponse`.

The allowlist is complete against today's 13 call sites, so this doesn't misfire now. The concern is maintenance: add a new `runCommand` site and forget the set, and the button just does nothing, with the only trace in the output channel.

Throwing instead gives the same defense with an observable failure — the bridge turns it into a rejected promise, which existing sites already handle (e.g. `BallerinaView.tsx:102` would surface its "Failed to start the setup process" message):

```ts
ext.logError(`Blocked disallowed webview command: ${props.command}`);
throw new Error(`Command not allowed: ${props.command}`);
```

Also worth noting in the `ALLOWED_WEBVIEW_COMMANDS` doc comment that this gates command **ids** only — `args` are still forwarded verbatim (`CloneProject`'s org/project, `git.push`), so a compromised webview retains control of the payloads.

wi/wi-extension/src/stateMachine.ts
gigara: Good catch — fixed in `2916a83`. Both sites now attach the error sink:

```ts
void context.webviewManager.showWelcome().catch((error) => ext.logError("Failed to show welcome webview", error as Error));
void context.webviewManager.show(view).catch((error) => ext.logError("Failed to show webview", error as Error));
```

This was a regression I introduced: making `show()` async to wait for the bridge port silently converted a propagating throw into a dropped rejection. `ext` was already imported in `stateMachine.ts`, so it's just the sink.

wi/wi-extension/src/ws-managers/main/ws-manager.ts
gigara: Agreed on all three — fixed in `2916a83`.

Now throws instead of returning a failed response:

```ts
// Forwarding arbitrary ids would make this a general command-execution
// primitive for anything that can reach the bridge. Throw rather than
// return a failed response: callers cannot detect the latter, since the
// success path resolves with whatever the command returned.
ext.logError(`Blocked disallowed webview command: ${props.command}`);
throw new Error(`Command not allowed: ${props.command}`);
```

I traced the chain you described before changing it, and it holds: `registerRoute` wraps the handler in try/catch and returns `createWsErrorResponse`; `WsClient.request` throws on `!response.success` (line 396); and `BallerinaView.tsx:102` already catches around `runCommand` and shows "Failed to start the setup process." So a blocked id now surfaces instead of doing nothing.

Your framing of the risk is the part that convinced me — the allowlist is complete today, so this isn't a live bug, but a silent no-op is the wrong failure mode for something whose whole purpose is to be maintained alongside a list.

Also added the `args` caveat to the doc comment:

```ts
 * Note this gates ids only — `args` are forwarded to the command as given, so a
 * compromised webview still controls the payloads of the commands listed here.
```

Worth being explicit that this is a real residual limitation, not just documentation: for `CloneProject` the org/project come from the webview, and `git.push` acts on the current repo. The allowlist bounds *which* commands can run, not what they're asked to do. Narrowing that would mean per-command argument validation, which I'd rather not fold into this PR — happy to raise it separately if you think it's warranted.

`wi-extension` typechecks with 0 errors after both changes.

NipunaRanasinghe (approved):
LGTM

PR #1992, opened 2026-08-03
Fix nightly Windows, extension-pin, and smoke-test failures

samithkavishke:
## Purpose

Three fixes for failures the nightly pipeline hit after `ci/versioning-and-nightly` landed. All of
them are CI-only: two workflows, one resolver script, one new helper, plus one stray file deletion.
Nothing in the product changes.

The commit that lets the nightly publish *while* smoke tests fail is deliberately **not** in this PR —
that is a temporary escape hatch and is tracked separately.

The concrete failures this addresses:

1. **Windows OpenSSL install breaks on a warm runner.** A persisted `C:\vcpkg\vcpkg.exe` is paired
   with freshly pulled scripts, and the old executable cannot parse the current
   `scripts/vcpkg-tools.json` schema.
2. **The nightly pins a Ballerina extension that cannot render the embedded creation form.** The
   rolling `nightly` tag keeps its name while the asset behind it changes, so filename matching
   selects a VSIX that is missing the federation remote. The build goes green and the smoke test
   fails in the UI with no indication of why.
3. **The upstream smoke-test suite still drives the old welcome page.** `wso2ipw@0.1.8` scripts click
   *Create New Integration*; the current welcome page says *Start your Integration Project*.

## Goals

Repair the nightly without changing how it is driven. No workflow is renamed and no existing input is
removed or retyped. `daily-build.yml` gains dispatch inputs (see below) but its scheduled behaviour is
byte-for-byte what it was.

## Approach

**3 commits, 6 files, +243/−47.**

### 1. `Fix nightly Windows and smoke build failures` (`b13f709`)

**`build.yml` — Windows OpenSSL.** `git pull` → `git pull --ff-only`, and `bootstrap-vcpkg.bat` now
runs unconditionally (with `-disableMetrics`) instead of only when `vcpkg.exe` is absent. The
repository and the executable have to move together; bootstrapping is idempotent, so the warm path
costs a rebuild and nothing else.

**`resolve-nightly-versions.sh` — reject an incompatible extension at pin time.** `select_release` is
split: `release_candidates()` now returns *every* eligible release (rolling nightly first, then
non-qualified releases newest-first) instead of just the head of the list. A new
`select_compatible_vsix_release()` walks that list, downloads each candidate VSIX and accepts the first
one whose archive contains a required entry.

`resolve_github_component` takes an optional 5th argument for that entry. Only
`ballerina.extension.version` passes one:

```
'extension/resources/jslibs/federation/remoteEntry.js'
```

Every other component keeps the cheap metadata-only path — no runtime distributions get downloaded
during resolution. Requires `unzip`, now noted in the script header.

**`smoke-test.yml` — fail loudly, and early.** A `Validate embedded creation form dependencies` step
(Linux + Windows) asserts the installed product actually ships
`extensions/wso2.ballerina/resources/jslibs/federation/remoteEntry.js` before any UI automation runs.
This is the belt to the resolver's braces: it also catches the case where the extension was pinned
correctly but never unpacked.

**`installers/linux-deb/package/DEBIAN/templates` — deleted.** The file's entire content was
`# No templates needed for WSO2 Integrator`. A `templates` file present in a DEBIAN control directory
is parsed by debconf; a comment-only one is not useful and the tooling is happier without it.

### 2. `Allow isolated manual nightly builds` (`a29dddc`)

`daily-build.yml` had no inputs, so the only way to exercise the pin-and-build flow was to dispatch it
and let it overwrite production nightly state. It now takes three:

| input | default | meaning |
| --- | --- | --- |
| `source_ref` | `main` | branch/tag/SHA to pin and build |
| `build_branch` | `builds/nightly-test` | force-updated branch recording the pinned commit |
| `publish_tag` | `nightly-test` | rolling release tag replaced on success |

The defaults are **test** names, so clicking "Run workflow" without thinking cannot touch production.

A `Validate build route` step guards the combinations:

- `build_branch` must live under `builds/` — nothing else will ever be force-pushed.
- `publish_tag` must be `nightly` or `nightly-*`.
- `build_branch` and `publish_tag` must differ, or the two become an ambiguous Git ref.
- `builds/nightly` and `nightly` are an all-or-nothing pair — you cannot aim a test branch at the
  production tag or vice versa.
- Targeting that production pair requires `source_ref == main`.

Scheduled runs bypass the inputs entirely (`github.event_name == 'schedule' && 'main' || inputs.…`),
so the production route stays an invariant regardless of how GitHub represents absent inputs. The
`concurrency` group is keyed on the destination branch, so test dispatches no longer serialise behind
the nightly.

### 3. `Fix smoke tests for current creation flow` (`de0fa3f`)

New `ci/build/patch-wso2ipw-smoke-tests.js`, run right after `npm install -g wso2ipw@0.1.8` on both
platforms, rewriting the old welcome-page heading to the current one in the two
`02-create-integration.sh` example scripts.

It is written to fail rather than silently no-op:

- throws if a script contains *neither* the old nor the current heading — the page changed again and
  someone needs to look;
- throws if any of four creation-form selectors it depends on has gone missing.

Both smoke jobs gain a `Checkout smoke-test support` step (`persist-credentials: false`) since they
previously never checked the repo out and now need this script.

**This is a stopgap.** The right fix is a `wso2ipw` release whose examples target the current UI; when
that ships, the patch script and both checkout steps come back out.

## Testing

- Dispatch **Daily Build** with `source_ref` = this branch, leaving `build_branch` and `publish_tag`
  at their test defaults. This exercises the full pin → build → publish path against
  `builds/nightly-test` / `nightly-test` without touching production.
- Confirm the resolver log shows `checking wso2/ballerina-vscode@… for
  extension/resources/jslibs/federation/remoteEntry.js`, and a `rejecting …` line if the rolling
  nightly asset is currently incomplete.
- Confirm `Validate embedded creation form dependencies` passes on both Linux and Windows.
- Confirm `Aligned …/02-create-integration.sh with the current welcome page` appears twice per job.
- Negative check on the route guard: dispatch with `build_branch: builds/nightly` and
  `publish_tag: nightly-test`; the run must fail in `Validate build route`.

## Notes for review

- The resolver now downloads candidate VSIXes during the pin step. In the normal case that is one
  download; the worst case is one per rejected release until a compatible one is found.
- `allow_smoke_test_failures` is **not** part of this PR.

## Summary by CodeRabbit

* **New Features**
  * Added configurable manual build runs with selectable source refs, build branches, and publishing tags.
  * Added compatibility checks to ensure required extension resources are included in release artifacts.
  * Added support for running smoke tests against a specified source revision.

* **Bug Fixes**
  * Improved Windows dependency setup and recovery when existing checkouts cannot be updated.
  * Updated smoke tests to verify the embedded creation form and align examples with the current welcome page.

* **Chores**
  * Improved nightly build coordination, version selection, and publishing reliability.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -93,6 +93,17 @@ jobs:
           echo "wso2-integrator payload binary not found under /usr/share/wso2-integrator or /opt/wso2-integrator" >&2
           exit 1
 
+      - name: Validate embedded creation form dependencies
+        run: |
+          PRODUCT_ROOT=$(dirname "$WSO2_INTEGRATOR_PATH")
+          remote="${PRODUCT_ROOT}/resources/app/extensions/wso2.ballerina/resources/jslibs/federation/remoteEntry.js"
+          if [ ! -f "${remote}" ]; then
+            echo "Error: wso2.ballerina does not provide the embedded creation form (${remote})." >&2
+            echo "       Reject this extension version before running the UI smoke test." >&2
+            exit 1
+          fi
+          echo "Found Ballerina creation form: ${remote}"
+
       - name: Patch ICP script for dash compatibility
         if: matrix.example == 'icp'
         run: |
@@ -234,6 +245,16 @@ jobs:
           Add-MpPreference -ExclusionPath $installDir
           Add-MpPreference -ExclusionPath "$env:USERPROFILE\WSO2Integrator"
           Add-MpPreference -ExclusionPath "$env:USERPROFILE\.ballerina"
+
+      - name: Validate embedded creation form dependencies
+        shell: pwsh
+        run: |
+          $productRoot = Split-Path $env:WSO2_INTEGRATOR_PATH
+          $remote = Join-Path $productRoot 'resources\app\extensions\wso2.ballerina\resources\jslibs\federation\remoteEntry.js'
+          if (-not (Test-Path $remote -PathType Leaf)) {
+            throw "wso2.ballerina does not provide the embedded creation form ($remote). Reject this extension version before running the UI smoke test."
+          }
+          Write-Host "Found Ballerina creation form: $remote"
 
       - name: Pre-pull ICP Ballerina modules
         if: matrix.example == 'icp'
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -13,6 +13,14 @@ on:
         required: false
         type: string
         default: ''
+      # This workflow now runs a script out of the repository, so it has to check out the commit that
+      # produced the artifact under test rather than whatever the caller happened to be triggered on.
+      # Empty means "whatever actions/checkout defaults to", which is the previous behaviour.
+      ref:
+        description: 'Commit to take the smoke-test support scripts from'
+        required: false
+        type: string
+        default: ''
   workflow_dispatch:
     inputs:
       integrator_version:
@@ -39,6 +47,7 @@ jobs:
       - name: Checkout smoke-test support
         uses: actions/checkout@v4
         with:
+          ref: ${{ inputs.ref }}
           persist-credentials: false
 
       - name: Install system dependencies
@@ -144,7 +153,7 @@ jobs:
         shell: bash
         run: |
           npm install -g wso2ipw@0.1.8
-          node ci/build/patch-wso2ipw-smoke-tests.js "$(npm root -g)/wso2ipw/examples"
+          node ci/build/patch-wso2ipw-smoke-tests.js "$(npm root -g)/wso2ipw/examples" "${{ matrix.example }}"
 
       - name: Install playwright-cli and Chrome
         if: matrix.example == 'icp'
@@ -221,6 +230,7 @@ jobs:
       - name: Checkout smoke-test support
         uses: actions/checkout@v4
         with:
+          ref: ${{ inputs.ref }}
           persist-credentials: false
 
       - name: Download build artifact
@@ -288,7 +298,7 @@ jobs:
         shell: bash
         run: |
           npm install -g wso2ipw@0.1.8
-          node ci/build/patch-wso2ipw-smoke-tests.js "$(npm root -g)/wso2ipw/examples"
+          node ci/build/patch-wso2ipw-smoke-tests.js "$(npm root -g)/wso2ipw/examples" "${{ matrix.example }}"
 
       - name: Install playwright-cli and Chrome
         if: matrix.example == 'icp'
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -36,6 +36,11 @@ jobs:
     env:
       DISPLAY: ':99'
     steps:
+      - name: Checkout smoke-test support
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
       - name: Install system dependencies
         run: |
           sudo apt-get update
@@ -136,7 +141,10 @@ jobs:
           node-version: '20'
 
       - name: Install wso2ipw
-        run: npm install -g wso2ipw@0.1.8
+        shell: bash
+        run: |
+          npm install -g wso2ipw@0.1.8
+          node ci/build/patch-wso2ipw-smoke-tests.js "$(npm root -g)/wso2ipw/examples"
 
       - name: Install playwright-cli and Chrome
         if: matrix.example == 'icp'
@@ -210,6 +218,11 @@ jobs:
       matrix:
         example: [hello-world-service, icp]
     steps:
+      - name: Checkout smoke-test support
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
       - name: Download build artifact
         if: inputs.windows_artifact_name != ''
         uses: actions/download-artifact@v4
@@ -272,7 +285,10 @@ jobs:
           node-version: '20'
 
       - name: Install wso2ipw
-        run: npm install -g wso2ipw@0.1.8
+        shell: bash
+        run: |
+          npm install -g wso2ipw@0.1.8
+          node ci/build/patch-wso2ipw-smoke-tests.js "$(npm root -g)/wso2ipw/examples"
 
       - name: Install playwright-cli and Chrome
         if: matrix.example == 'icp'
--- a/ci/build/patch-wso2ipw-smoke-tests.js
+++ b/ci/build/patch-wso2ipw-smoke-tests.js
@@ -1,9 +1,13 @@
 const fs = require('fs');
 const path = require('path');
 
-const examplesRoot = process.argv[2];
-if (!examplesRoot) {
-  throw new Error('Usage: node patch-wso2ipw-smoke-tests.js <wso2ipw-examples-directory>');
+// The example comes from the caller rather than a list in here: the workflow's `example` matrix is the
+// single source of truth for which ones exist. Hardcoding them here would let a new matrix entry go
+// unpatched — the checks below only fire for names this script is told about, so that example would
+// drive the old welcome page and fail in the UI instead of here.
+const [, , examplesRoot, example] = process.argv;
+if (!examplesRoot || !example) {
+  throw new Error('Usage: node patch-wso2ipw-smoke-tests.js <wso2ipw-examples-directory> <example>');
 }
 
 const oldHeading = 'Create New Integration';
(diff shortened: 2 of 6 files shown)


ci/build/resolve-nightly-versions.sh
```diff
@@ -91,23 +93,93 @@ select_release() {
...
+    fi
+
+    if ! entries=$(unzip -Z1 "${archive}"); then
+      rm -rf "${check_dir}"
+      echo "Error: ${asset} from ${repo}@${tag} is not a readable zip archive." >&2
+      return 2
+    fi
+    rm -rf "${check_dir}"
+
+    if printf '%s\n' "${entries}" | grep -Fxq "${required_entry}"; then
```
NipunaRanasinghe: **This rejects compatible VSIXes. `grep -q` in a pipe under `pipefail` — reproduced locally.**

The script sets `set -euo pipefail` (line 13). `grep -q` exits immediately on first match, so `printf` takes SIGPIPE and the pipeline reports 141 — the `if` goes **false on a successful match**:

```
after-match:    500 lines /    25899 bytes | pipe: match        herestring: match
after-match:   1000 lines /    51900 bytes | pipe: match        herestring: match
after-match:   1500 lines /    78400 bytes | pipe: MISS(rc=141) herestring: match
after-match:  40000 lines /  2148901 bytes | pipe: MISS(rc=141) herestring: match
```

It triggers once more than ~64KB of listing follows the matched entry (the pipe buffer), so it is **layout-dependent**: whether a compatible VSIX is accepted depends on where `remoteEntry.js` sits in that archive's central directory, which shifts as the extension changes.

The failure mode is the bad one. `select_compatible_vsix_release` walks candidates in order, and each one greps independently — so this can reject the correct newest release and then accept an older one whose entry happens to fall within 64KB of the end. That is precisely the "silently pin an older extension" outcome this function's own header comment says it exists to prevent.

You documented this exact trap in [ballerina-vscode#763](https://github.com/ballerina-platform/ballerina-vscode/pull/763):

> Listed once into a variable: piping unzip into 'grep -q' lets grep exit on the first match, and the resulting SIGPIPE fails the pipeline under pipefail.

Same fix, which I confirmed is immune at every size tested:

```suggestion
    if grep -Fxq "${required_entry}" <<<"${entries}"; then
```

To be straight about the evidence: I have nothing showing this is firing today, and nothing showing it isn't — the resolver never runs in this PR's CI (see review body).

ci/build/resolve-nightly-versions.sh
```diff
@@ -91,23 +93,93 @@ select_release() {
              asset: ([.assets[]? | select(.name | test(\"${jq_regex}\")) | .name][0])}
           | select(.asset != null) ]
     | (map(select(.nightly))[0] // empty | \"nightly\t\(.tag)\t\(.asset)\"),
-      (map(select(.qualified == false))[0] // empty | \"latest\t\(.tag)\t\(.asset)\")
+      (map(select(.nightly == false and .qualified == false))[] | \"latest\t\(.tag)\t\(.asset)\")
```
NipunaRanasinghe, line 96: **The fallback walk is unbounded — worth capping.**

Changing `[0]` to `[]` means every eligible non-nightly release is now a candidate, up to the `per_page=100` ceiling. Combined with `select_compatible_vsix_release` downloading each one, a bad rolling nightly costs up to ~100 large VSIX downloads before the resolver gives up.

The federation remote is brand new — it is what ballerina-vscode#763 adds — so releases predating that packaging fix can never satisfy the contract. Walking past the first few is guaranteed-wasted time and rate limit.

Suggest capping the inspected candidates (3–5) and failing with a message that says the cap was hit, so the log distinguishes "checked everything plausible" from "gave up early". "Notes for review" says *"the worst case is one per rejected release"* without naming what that total is, which undersells it.

ci/build/resolve-nightly-versions.sh
```diff
@@ -162,7 +234,8 @@ echo "Resolving nightly component versions into ${VERSIONS_FILE}"
 echo "GitHub-release components:"
 
 resolve_github_component "ballerina.extension.version" \
-  "wso2/ballerina-vscode" '^ballerina-(.+)\.vsix$' "v"
+  "wso2/ballerina-vscode" '^ballerina-(.+)\.vsix$' "v" \
+  'extension/resources/jslibs/federation/remoteEntry.js'
```
NipunaRanasinghe, line 261: **Merge order matters here, and the description doesn't say so.**

This contract cannot be satisfied by any currently published `wso2/ballerina-vscode` release — shipping `remoteEntry.js` in the VSIX is exactly what ballerina-vscode#763 fixes, and that PR is still open.

So if this merges first, the next nightly rejects the rolling `nightly` asset, downloads and rejects the entire fallback list, then `exit 1`. The nightly fails outright rather than degrading to the previous behaviour.

Please note the prerequisite explicitly in the description: ballerina-vscode#763 merged **and** one successful nightly run publishing a VSIX that contains this entry. Otherwise whoever merges this first inherits a red nightly with a confusing "no compatible release" error.

.github/workflows/smoke-test.yml
```diff
@@ -36,6 +36,11 @@ jobs:
     env:
       DISPLAY: ':99'
     steps:
+      - name: Checkout smoke-test support
```
NipunaRanasinghe, line 47: **Unpinned checkout — this undercuts the isolated-test route from commit 2.**

No `ref:` here, and `build.yml` calls this workflow passing only `artifact_name` / `windows_artifact_name`. So `patch-wso2ipw-smoke-tests.js` comes from the caller's triggering ref, not from the commit being built. `build.yml:50` already documents the mechanism:

> One thing this cannot fix: with `uses: ./.github/workflows/...` the reusable workflow *definitions* are always read from the caller's ref, never from this input. That is harmless here, since both dispatchers cut their branch from the ref they were dispatched on and add nothing but a version commit.

That justification held while this workflow never checked the repo out. It now checks out and **executes a script from it**, so the reasoning needs revisiting.

Concretely: dispatch Daily Build **from `main`** with `source_ref: ci/nightly-build-fixes` — the smoke test checks out `main`, where this script does not exist yet, and `node ci/build/patch-wso2ipw-smoke-tests.js` fails. That is the exact flow commit 2 exists to enable, and it only works if you dispatch *from* the branch rather than targeting it via `source_ref`.

Either thread a `ref` input through `build.yml` into this workflow, or state the dispatch-from-the-branch constraint in the Testing section — the current step 1 reads as though either works.

.github/workflows/smoke-test.yml
```diff
@@ -251,7 +285,10 @@ jobs:
           node-version: '20'
 
       - name: Install wso2ipw
-        run: npm install -g wso2ipw@0.1.8
+        shell: bash
+        run: |
+          npm install -g wso2ipw@0.1.8
+          node ci/build/patch-wso2ipw-smoke-tests.js "$(npm root -g)/wso2ipw/examples"
```
NipunaRanasinghe: Nit: this is the first `shell: bash` step in the Windows job, and `$(npm root -g)` returns a backslash path (`C:\Users\runneradmin\AppData\Roaming\npm\node_modules`).

It should be fine — bash double quotes preserve backslashes literally, and Node's `path.join` normalises mixed separators — but it is worth eyeballing on a real Windows run rather than assuming, since a silent failure here means the smoke test drives the old welcome page and fails later in the UI with a much less obvious cause.

ci/build/patch-wso2ipw-smoke-tests.js
```diff
@@ -0,0 +1,35 @@
...
+const oldHeading = 'Create New Integration';
+const currentHeading = 'Start your Integration Project';
+const requiredCreationFormSelectors = [
+  "getByRole('button', {name: 'Create', exact: true})",
+  'Integration Name',
+  'Project Name',
+  'Create Integration',
+];
+
+for (const example of ['hello-world-service', 'icp']) {
```
NipunaRanasinghe: Nit: this duplicates the workflow matrix (`example: [hello-world-service, icp]`) in a second place. If the matrix gains an example, this script silently won't patch it — the throws only fire for the two names hardcoded here, so the new example would drive the old welcome page and fail in the UI.

Deriving the list from `readdirSync(examplesRoot)`, or passing the example name as a second argv from the matrix, keeps the two in step. Given the file is explicitly a stopgap this may not be worth it — but if it outlives the `wso2ipw` fix, this is where it will bite.

.github/workflows/smoke-test.yml
samithkavishke: Taken. `smoke-test.yml` now has an optional `ref` input used by both `Checkout smoke-test support` steps, and `build.yml` passes `ref: ${{ inputs.ref }}` to both smoke jobs the same way `compile` already received.

.github/workflows/smoke-test.yml
samithkavishke: This isn't new so we will leave it as it is. More to the point, the failure mode you're guarding against can't occur here. If the path were wrong the script wouldn't silently no-op into a confusing UI failure — `fs.readFileSync` throws `ENOENT` and the step dies immediately. Confirmed locally:

ci/build/patch-wso2ipw-smoke-tests.js
samithkavishke: Addressed

NipunaRanasinghe (approved):
LGTM

PR #2014, opened 2026-08-04
Fix nightly build rolling issue and parameterize the smoke tests for the release

samithkavishke:
## Purpose
$Subject; currently nightly tag is not overridden but it currently it pushes the new nightly build and since it has a new timestamp it is not overriding but it put it under the same tag

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved nightly release publishing reliability by verifying uploaded assets before updating the release.
  * Preserved the previous nightly release when publication is incomplete, reducing disruption from partial uploads.
  * Ensured nightly release tags point to the intended changes after successful publication.
  * Removed only superseded assets after successful publication while retaining valid existing assets.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1264,8 +1264,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      # `nightly` is a rolling tag and every night's filenames carry a fresh timestamp, so the
-      # upload above adds to the previous night's assets instead of replacing them. They are cleared
+      # A `nightly*` tag is rolling, and every build's filenames carry a fresh timestamp, so the
+      # upload above adds to the previous build's assets instead of replacing them. They are cleared
       # here rather than before the upload: the tag is what external consumers pin their download
       # URLs to, and deleting first would leave it empty — or, if any step in between failed, gone —
       # so a complete set is always downloadable. Everything this build produced is verified present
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -1278,3 +1263,59 @@ jobs:
...
+      # assets at all.
+      #
+      # The tag is moved for the same reason it is not deleted: `action-gh-release` only creates a
+      # tag that does not yet exist, so without this the rolling tag would stay pinned to whichever
+      # commit first published it while its assets moved on.
+      #
+      # GH_REPO is required, not belt-and-braces: this job never checks the repository out, so `gh`
+      # has no remote to infer the repository from and every call fails without it.
+      - name: Roll the nightly tag and drop superseded assets
+        if: ${{ inputs.publish_tag == 'nightly' }}
```
pasindufernando1: **This gate makes the step impossible to dry-run, and it hasn't been run yet.**

There are no workflow runs on `nightlyBuildRolling` and no `nightly` release on the fork, so this code has never executed anywhere. Its first execution will be against the live rolling tag, where it will delete the 44 assets currently attached.

The blast radius is asymmetric, which is worth keeping in mind: a bug in the verify phase just exits 1 without touching anything, but a bug in the `superseded` computation over-deletes. So it's specifically the delete loop that wants exercising once somewhere disposable — a fork release seeded with a couple of dummy assets is enough to prove it.

The obstacle is this line. `daily-build.yml` accepts `nightly` *or* any `nightly-` prefixed tag, so a `nightly-test` dispatch would be the natural dry run — but it skips this step entirely. Two options:

- Relax the gate to `startsWith(inputs.publish_tag, 'nightly')` and use `${{ inputs.publish_tag }}` in place of the four hardcoded `nightly` literals below (lines 1297, 1307, 1319 and the `git/refs/tags/nightly` path). That makes the path permanently dry-runnable, not just once.
- Or leave the gate as-is and run it manually on the fork with the real tag name before merging.

Either is fine — but given this deletes release assets and has never run, I'd rather not have the first execution be the production nightly at 06:30 UTC.

.github/workflows/build.yml
```diff
@@ -1278,3 +1263,72 @@ jobs:
...
+          fi
+          attached=$(gh release view "${PUBLISH_TAG}" --json assets --jq '.assets[].name' \
+            | sed '/^$/d' | sort)
+
+          # An upload that silently dropped a file would otherwise be "fixed" by deleting the
+          # previous, complete set. Bail out and leave both sets attached instead.
+          missing=$(comm -23 <(printf '%s\n' "${published}") <(printf '%s\n' "${attached}"))
+          if [ -n "${missing}" ]; then
+            echo "Error: these files did not reach the ${PUBLISH_TAG} release:" >&2
+            printf '  - %s\n' ${missing} >&2
```
NipunaRanasinghe: `${missing}` is unquoted here, so the per-line listing relies on word-splitting the newline-joined string back into separate `printf` arguments. That only works because release filenames don't contain spaces/tabs — harmless today, but fragile. Doesn't affect safety (the script still `exit 1`s either way, keeping the previous assets), just the diagnostic's formatting.

Same fix already used a few lines down for `superseded` would make this robust too:
```bash
while IFS= read -r name; do
  echo "  - ${name}"
done <<< "${missing}" >&2
```

NipunaRanasinghe (dismissed):
Approved with a non-blocking comment

.github/workflows/build.yml
```diff
@@ -1278,3 +1263,72 @@ jobs:
...
+      # alone would make the destructive branch unreachable outside production. It stays false for
+      # the other two callers: release.yml passes `v<version>` and dev-build.yml passes ''.
+      #
+      # PUBLISH_TAG is bound through `env` rather than interpolated into the script for the reason
+      # release.yml spells out: `${{ }}` is expanded before bash parses the body, so a value carrying
+      # shell syntax would change the commands being run rather than the tag they act on.
+      #
+      # GH_REPO is required, not belt-and-braces: this job never checks the repository out, so `gh`
+      # has no remote to infer the repository from and every call fails without it. That is exactly
+      # what silently disabled the delete-first step this replaces.
```
NipunaRanasinghe: Above code comment looks too verbose. Better to make it concise.

.github/workflows/build.yml
samithkavishke: Addressed

.github/workflows/build.yml
samithkavishke: Addressed

NipunaRanasinghe (approved):

PR #2067, opened 2026-08-10
Nightly build: auto-select staging/* source branch and release nightly on test case failure

samithkavishke:
## Why

Development is moving from `main` onto a `staging/<version>` branch (e.g. `staging/5.1.0`) that
carries `integrator.version=5.1.0-alpha` instead of `-SNAPSHOT`. Two things needed to change for
the nightly build to keep working correctly once that branch is the active line:

1. The scheduled nightly hardcoded its source ref to `main`. It now looks for a `staging/*` branch
   on the remote first, and only falls back to `main` if none exists.
2. `apply-version.sh` only knew how to strip-and-timestamp a `-SNAPSHOT` suffix. Left alone, a
   nightly built from `integrator.version=5.1.0-alpha` would publish that exact, un-timestamped
   string every night, silently overwriting the rolling `nightly` release with an identical
   version each run.

## What changed

- `.github/workflows/daily-build.yml`:
  - On `schedule` events only, resolves `SOURCE_REF` by querying
    `repos/{repo}/git/matching-refs/heads/staging/` via `gh api` (no checkout has happened yet at
    that point in the job, so `git ls-remote` against a configured `origin` isn't available).
    Zero matches keeps `main`; more than one match fails the run loudly instead of guessing which
    to build. `workflow_dispatch` is untouched — it always uses its explicit `source_ref` input.
  - The production-branch safety guard now accepts `main` or any `staging/*` ref (previously
    `main` only).
  - The "declares SNAPSHOT" assertion is renamed/generalized: `integrator.version` may now also
    carry any alphabetic pre-release qualifier (`-alpha`, `-alpha2`, `-beta`, `-rc1`, ...), not just
    `-alpha`; `wi.extension.version` is unchanged and stays `-SNAPSHOT`-only.
- `ci/build/apply-version.sh`: the product-version shape check now matches `-[A-Za-z]+[0-9]*$`
  instead of only `-SNAPSHOT`, producing `5.1.0-<yyyymmddHHmm>` for any such qualifier — same shape
  as `-SNAPSHOT` today, qualifier dropped rather than appended after the timestamp. An
  already-concrete, already-stamped version (e.g. `5.1.0-202608101200`) still passes through
  unchanged, since it has no leading letter for the regex to match.

`ci/build/apply-version.sh` also needs this same fix on `staging/5.1.0` itself, since that script
runs from whichever branch gets checked out at build time — that's a separate PR
(#TBD, targeting `staging/5.1.0`).

## Test plan

- [x] `bash -n` on the modified script and all `run:` blocks in the workflow — syntax OK
- [x] YAML parses cleanly
- [x] Dry-ran `apply-version.sh` against scratch copies of `component-versions.properties` for
      `5.1.0-alpha`, `5.1.0-beta`, `5.1.0-rc1`, `5.1.0-SNAPSHOT` (all → `5.1.0-<timestamp>`) and
      `5.1.0-202608101200` (already concrete → passes through unchanged)
- [x] Confirmed `gh api .../matching-refs/heads/staging/` returns the expected `staging/5.1.0` ref
      against the live repo, and `[]` for a nonexistent prefix
- [ ] Exercise via `workflow_dispatch` with a test `build_branch`/`publish_tag` before relying on
      the scheduled run

## Summary by CodeRabbit

* **Build and Release Improvements**
  * Scheduled builds now automatically use the appropriate staging branch when exactly one is available.
  * Builds remain on the main branch when no staging branch exists and clearly fail when multiple staging branches are detected.
  * Production releases now support sources from the main branch or staging branches.
  * Version handling now supports alphabetic pre-release labels such as alpha, beta, and release candidates.
  * Snapshot version requirements remain enforced for extension releases.

### DIFF
--- a/.github/workflows/daily-build.yml
+++ b/.github/workflows/daily-build.yml
@@ -42,12 +42,12 @@ on:
         description: 'Force-updated branch that records the pinned build commit'
         required: true
         type: string
-        default: builds/nightly-test
+        default: builds/nightly
       publish_tag:
         description: 'Rolling release tag to replace after every job succeeds'
         required: true
         type: string
-        default: nightly-test
+        default: nightly
       gate_on_smoke_tests:
         description: 'Fail the run and withhold the release if a smoke test fails. Ignored on the scheduled production run, which never gates on them.'
         required: true
--- a/.github/workflows/daily-build.yml
+++ b/.github/workflows/daily-build.yml
@@ -4,8 +4,8 @@ name: Daily Build
 #
 #   1. resets the `builds/nightly` branch to `main`, or to a `staging/<version>` branch if exactly
 #      one exists on the remote
-#   2. pins every dependency to its newest nightly build, or its newest GA release when the upstream
-#      repo publishes none
+#   2. pins each tracked dependency to its newest nightly build, or its newest release when the
+#      upstream repo publishes none
 #   3. stamps the timestamped product and extension versions and commits, so
 #      `git diff <source> builds/nightly` is exactly the pin and the build is reproducible from that
 #      commit
--- a/.github/workflows/daily-build.yml
+++ b/.github/workflows/daily-build.yml
@@ -49,11 +49,11 @@ on:
         required: true
         type: string
         default: nightly-test
-      run_smoke_tests:
-        description: 'Whether to run the UI smoke tests. Ignored on the scheduled production run, which always runs them.'
+      gate_on_smoke_tests:
+        description: 'Fail the run and withhold the release if a smoke test fails. Ignored on the scheduled production run, which never gates on them.'
         required: true
         type: boolean
-        default: true
+        default: false
 
 permissions:
   id-token: write
@@ -280,8 +280,7 @@ jobs:
       build_macos: true
       build_windows: true
       build_packed_installers: true
-      # Ignore the dispatch input on a scheduled run, same as the route env vars above: the
-      # production nightly always exercises the smoke tests.
-      run_smoke_tests: ${{ github.event_name == 'schedule' && true || inputs.run_smoke_tests }}
+      run_smoke_tests: true
+      smoke_tests_required: ${{ github.event_name != 'schedule' && inputs.gate_on_smoke_tests == true }}
       ballerina_extension_source: github
       mi_extension_source: github
--- a/.github/workflows/daily-build.yml
+++ b/.github/workflows/daily-build.yml
@@ -130,11 +130,17 @@ jobs:
           git check-ref-format --branch "${BUILD_BRANCH}" >/dev/null
           git check-ref-format "refs/tags/${PUBLISH_TAG}" >/dev/null
 
-          # The only remaining restriction: build_branch is force-pushed verbatim, so a typo here
-          # would overwrite whatever branch was named.
+          # build_branch is force-pushed verbatim, so a typo here would overwrite whatever branch
+          # was named.
           case "${BUILD_BRANCH}" in
             builds/*) ;;
             *) echo "Error: build_branch must be below builds/; refusing to force-push ${BUILD_BRANCH}." >&2
+               exit 1 ;;
+          esac
+          # publish_tag's release is deleted and recreated, so it must never name a GA release.
+          case "${PUBLISH_TAG}" in
+            nightly|nightly-*) ;;
+            *) echo "Error: publish_tag must be nightly or start with nightly-." >&2
                exit 1 ;;
           esac
 
--- a/.github/workflows/daily-build.yml
+++ b/.github/workflows/daily-build.yml
@@ -173,12 +173,12 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
 
       # apply-version.sh is shape-based: it only rewrites `-SNAPSHOT` (or, on integrator.version
-      # only, an `-alpha`/`-alphaN` qualifier), and passes a concrete version through verbatim. That
-      # is what makes it safe to run everywhere, but it also means a source ref left on a concrete
-      # version after a release (the manual bump documented in release.yml) would make every nightly
-      # publish that same un-timestamped version indefinitely — and because the rolling release is
-      # deleted and recreated each run, it would still look like it worked. Fail before anything is
-      # pushed instead.
+      # only, any other alphabetic pre-release qualifier like `-alpha`/`-alpha2`/`-beta`/`-rc1`), and
+      # passes a concrete version through verbatim. That is what makes it safe to run everywhere,
+      # but it also means a source ref left on a concrete version after a release (the manual bump
+      # documented in release.yml) would make every nightly publish that same un-timestamped version
+      # indefinitely — and because the rolling release is deleted and recreated each run, it would
+      # still look like it worked. Fail before anything is pushed instead.
       - name: Assert source ref still declares pre-release versions
         shell: bash
         run: |
@@ -193,7 +193,7 @@ jobs:
             # it and stamps a timestamp exactly as it does for -SNAPSHOT. wi.extension.version has
             # no equivalent and stays -SNAPSHOT-only.
             if [ "${key}" = "integrator.version" ]; then
-              allowed='-(SNAPSHOT|alpha[0-9]*)$'
+              allowed='-[A-Za-z]+[0-9]*$'
             else
               allowed='-SNAPSHOT$'
             fi
--- a/.github/workflows/daily-build.yml
+++ b/.github/workflows/daily-build.yml
@@ -181,15 +181,14 @@ jobs:
           ref: ${{ steps.configure.outputs.source_ref }}
           token: ${{ secrets.GITHUB_TOKEN }}
 
-      # apply-version.sh re-stamps integrator.version only when it carries a pre-release qualifier
-      # (-SNAPSHOT/-alpha/-beta/-rc1/...); a concrete version passes through verbatim, so a missed
-      # post-release bump would make every nightly republish it. wi.extension.version needs no such
-      # check — the stamp step below forces it.
+      # The stamp step below would timestamp a bare `5.1.0` too, so this is not what makes the
+      # nightly safe — it is a tripwire: a source ref without a qualifier means a post-release bump
+      # was missed, and that should be seen rather than silently absorbed.
       - name: Assert source ref still declares a pre-release product version
         shell: bash
         run: |
           key=integrator.version
-          allowed='-[A-Za-z]+[0-9]*$'
+          allowed='-[A-Za-z][A-Za-z0-9.]*$'
           value=$(awk -F= -v k="${key}" '$1 == k { print substr($0, index($0, "=") + 1); exit }' \
             ci/build/component-versions.properties | tr -d '\r')
           if [ -z "${value}" ]; then
@@ -219,7 +218,7 @@ jobs:
         id: stamp
         run: |
           chmod +x ./ci/build/apply-version.sh
-          ./ci/build/apply-version.sh --force-extension
+          ./ci/build/apply-version.sh --force-product --force-extension
 
       - name: Upload versions
         uses: actions/upload-artifact@v4
--- a/.github/workflows/daily-build.yml
+++ b/.github/workflows/daily-build.yml
@@ -30,6 +30,7 @@ name: Daily Build
 
 on:
   schedule:
+    # 21:30 UTC = 03:00 AM IST the next day.
     - cron: '30 21 * * *'
   workflow_dispatch:
     inputs:
@@ -52,7 +53,7 @@ on:
         description: 'Fail the run and withhold the release if a smoke test fails. Ignored on the scheduled production run, which never gates on them.'
         required: true
         type: boolean
-        default: false
+        default: true
 
 permissions:
   id-token: write
@@ -119,9 +120,9 @@ jobs:
             else
               SOURCE_REF="${staging_candidates[0]}"
               echo "staging/* branch found; scheduled run uses ${SOURCE_REF}."
-              # job-level `env:` above was evaluated before this step ran and cannot see this
-              # value; every later step in this job reads $SOURCE_REF from the environment, so it
-              # must be re-exported here for them to see the resolved branch too.
+              # The checkout and the route output read steps.configure.outputs.source_ref, not this.
+              # Re-exported only so the later log and error lines that read $SOURCE_REF name the
+              # resolved branch instead of the job-level default.
               echo "SOURCE_REF=${SOURCE_REF}" >> "$GITHUB_ENV"
             fi
           fi
--- a/.github/workflows/daily-build.yml
+++ b/.github/workflows/daily-build.yml
@@ -97,10 +97,20 @@ jobs:
           # rather than `git ls-remote origin`, since no checkout — and so no `origin` remote —
           # exists yet at this point in the job.
           if [ "${GITHUB_EVENT_NAME}" = "schedule" ]; then
-            mapfile -t staging_candidates < <(
+            # Captured outside the `mapfile` process substitution: a subshell's exit status is
+            # invisible to the parent, so `gh api` failing (outage, permissions) would otherwise
+            # look identical to "no staging branches" and silently fall back to main.
+            if ! staging_refs=$(
               gh api "repos/${GITHUB_REPOSITORY}/git/matching-refs/heads/staging/" --jq '.[].ref' \
                 | sed 's#^refs/heads/##'
-            )
+            ); then
+              echo "Error: could not list staging/* branches." >&2
+              exit 1
+            fi
+            staging_candidates=()
+            if [ -n "${staging_refs}" ]; then
+              mapfile -t staging_candidates <<< "${staging_refs}"
+            fi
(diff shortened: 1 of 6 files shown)


.github/workflows/daily-build.yml
```diff
@@ -139,27 +182,39 @@ jobs:
...
+              exit 1
+            fi
+            # integrator.version alone may also carry an -alpha[N] marker: apply-version.sh strips
+            # it and stamps a timestamp exactly as it does for -SNAPSHOT. wi.extension.version has
+            # no equivalent and stays -SNAPSHOT-only.
+            if [ "${key}" = "integrator.version" ]; then
+              allowed='-[A-Za-z]+[0-9]*$'
+            else
+              allowed='-SNAPSHOT$'
+            fi
```
NipunaRanasinghe: **Blocker: this makes the scheduled nightly fail every night.**

`staging/5.1.0` currently declares:

```
integrator.version=5.1.0-alpha
wi.extension.version=1.1.26081017     # already concrete, not -SNAPSHOT
```

Running this loop's logic against that actual file:

```
PASS  integrator.version=5.1.0-alpha
FAIL  wi.extension.version=1.1.26081017  (expected match -SNAPSHOT$)
```

So the first scheduled run resolves `SOURCE_REF=staging/5.1.0`, checks it out, and exits 1 right here. It fails before the `git push` in the next step, so nothing gets corrupted — but the nightly stays dead. The PR description lists "`wi.extension.version` is unchanged and stays `-SNAPSHOT`-only" as a no-op; against this branch it's the failure.

**Loosening this regex alone isn't enough.** Dry-running `apply-version.sh` against the real staging properties file:

```
integrator.version=5.1.0-202608140957 (was 5.1.0-alpha)     # correct
wi.extension.version=1.1.26081017 (was 1.1.26081017)        # frozen
```

`apply-version.sh:147` only restamps the extension on `*-SNAPSHOT` or `--force-extension`, so a concrete `1.1.26081017` passes through verbatim and every nightly publishes an identical extension version — exactly the "silently overwriting the rolling `nightly` release with an identical version each run" failure this PR is about, just moved from the product to the extension. And per the comment two lines above the assert, the deleted-and-recreated rolling release means it would still look like it worked.

The script already anticipates the fix: `apply-version.sh:148-149` notes an already-odd minor is kept "so a derived version can be committed in place and re-derived next time" — which is what `--force-extension` exists for. The nightly's stamp step at line 234 calls the script bare. So it looks like two changes are needed:

1. `./ci/build/apply-version.sh --force-extension` in the stamp step, and
2. this guard accepting an already-odd-minor concrete extension version (the repeatable pre-release shape) alongside `-SNAPSHOT`, rather than `-SNAPSHOT` only.

Worth confirming against a `workflow_dispatch` run with `source_ref: staging/5.1.0` before merge — that's the unchecked box in the test plan, and it's the one that would have caught this.

ci/build/apply-version.sh
```diff
@@ -109,7 +112,7 @@ if [ -n "${EXPLICIT_VERSION}" ]; then
   fi
   require_semver "$(base_of "${EXPLICIT_VERSION}")" "--version"
   PRODUCT_VERSION="${EXPLICIT_VERSION}"
-elif [[ "${DECLARED_PRODUCT}" == *"-SNAPSHOT" ]]; then
+elif [[ "${DECLARED_PRODUCT}" =~ -[A-Za-z]+[0-9]*$ ]]; then
```
NipunaRanasinghe: Minor, and not reachable from CI today: this widened condition also changes `--mode release`.

With `DECLARED_PRODUCT=5.1.0-alpha` and `--mode release`, the branch below now yields `PRODUCT_VERSION=5.1.0` — a GA version. Before this change `5.1.0-alpha` fell through to the `else` and passed through verbatim.

CI can't hit it: `release.yml:162` always pairs `--mode release` with `--version`, and the `-n "${EXPLICIT_VERSION}"` test at line 108 short-circuits first. But someone running `./ci/build/apply-version.sh --mode release` by hand on a staging branch would now quietly get `5.1.0` where they used to get `5.1.0-alpha`. Given this script is described as "the one place a product or extension version is ever derived or written", an explicit guard seems worth it — or at least don't leave it implicit.

Either way the usage doc at lines 21-22 needs updating: it still says `release` "strips -SNAPSHOT instead of timestamping", which is no longer the full set of qualifiers it strips.

.github/workflows/daily-build.yml
```diff
@@ -87,6 +92,41 @@ jobs:
...
+              echo "Error: more than one staging/* branch exists; refusing to guess which to build." >&2
+              printf '  %s\n' "${staging_candidates[@]}" >&2
+              exit 1
+            else
+              SOURCE_REF="${staging_candidates[0]}"
+              echo "staging/* branch found; scheduled run uses ${SOURCE_REF}."
+              # job-level `env:` above was evaluated before this step ran and cannot see this
+              # value; every later step in this job reads $SOURCE_REF from the environment, so it
+              # must be re-exported here for them to see the resolved branch too.
+              echo "SOURCE_REF=${SOURCE_REF}" >> "$GITHUB_ENV"
```
NipunaRanasinghe: Nit on the comment rather than the code: "every later step in this job reads `$SOURCE_REF` from the environment" overstates the dependency.

The two things that actually matter — the checkout at line 182 and the route output — go through `steps.configure.outputs.source_ref`, set from the shell variable at line 171. The only later reads of `$SOURCE_REF` are the error text at line 213 and the echo at line 223, both log-only.

That matters because whether a `GITHUB_ENV` write actually overrides an entry already present in job-level `env:` isn't something I'd assert either way. Keeping the export is fine and the blast radius if precedence goes the other way is one wrong branch name in a log line — but the comment currently reads as though the build depends on it, which would send the next reader looking for a correctness bug that isn't there.

samithkavishke: Resolved

.github/workflows/daily-build.yml
```diff
@@ -38,17 +42,17 @@ on:
         description: 'Force-updated branch that records the pinned build commit'
         required: true
         type: string
-        default: builds/nightly-test
+        default: builds/nightly
```
NipunaRanasinghe, line 46: **The production guards were deleted, not relaxed — and the defaults now point at production.**

The description says "The production-branch safety guard now accepts `main` or any `staging/*` ref (previously `main` only)". Reading the diff, three checks are removed outright: `build_branch != publish_tag`, the paired `builds/nightly`⇔`nightly` destination check, and `production_branch ⇒ SOURCE_REF == main`. Nothing replaces them.

In the same hunk `build_branch` defaults `builds/nightly-test → builds/nightly` and `publish_tag` `nightly-test → nightly`. So: open "Run workflow", type any `source_ref`, leave the rest at default → `builds/nightly` is force-pushed to that branch, and the `nightly` release external consumers pin download URLs to is deleted and recreated from that build. Every one of those steps was rejected before this PR.

If relaxing really was the intent, the guard survives as:

```bash
if [ "${production_branch}" = "true" ]; then
  case "${SOURCE_REF}" in
    main|staging/*) ;;
    *) echo "Error: production nightly destinations may only be built from main or staging/*." >&2
       exit 1 ;;
  esac
fi
```

If the deletion is deliberate, that's your call to make — but it needs to be in the description instead of the sentence that says the opposite, because the removed guarantee is the one the header comment and the old `publish_tag` note both still lean on.

(The `build_branch != publish_tag` removal on its own is fine — `builds/*` and `nightly|nightly-*` are disjoint, so they can't collide.)

ci/build/resolve-nightly-versions.sh
```diff
@@ -1,11 +1,16 @@
...
+# repo publishes no nightly) and pin the result into component-versions.properties.
 #
 # It runs on the `builds/nightly` branch before the build is dispatched, so the build itself just
 # reads the committed file — which also makes every nightly reproducible from the commit this leaves
 # behind.
-# The product and extension versions are not touched here; ci/build/apply-version.sh owns those.
+#
+# Not touched here, on purpose:
+#   integrator.version, wi.extension.version   ci/build/apply-version.sh owns those
+#   ballerina.jre.version                      hand-maintained: the JRE has to match the distribution
```
NipunaRanasinghe: This says `ballerina.jre.version` is hand-maintained and not touched here. Line 272 still resolves it:

```bash
resolve_github_component "ballerina.jre.version" \
  "ballerina-platform/ballerina-custom-jre" '^ballerina-jre-linux-64-(.+)\.zip$' ""
```

So every nightly overwrites it, resolved independently of `ballerina.version` — which is what the comment says must not happen. Together with the dropped qualifier filter, the nightly can pin `ballerina.version` to an RC and `ballerina.jre.version` to an unrelated newest JRE release: exactly the mismatch this paragraph claims is prevented.

Either drop the `resolve_github_component` call, or drop the bullet. A wrong comment in the file that decides what ships in the runtime is worse than none.

.github/workflows/daily-build.yml
```diff
@@ -139,28 +158,28 @@ jobs:
...
-      # through verbatim. That is what makes it safe to run everywhere, but it also means a `main`
-      # left on a concrete version after a release (the manual bump documented in release.yml) would
-      # make every nightly publish that same un-timestamped version indefinitely — and because the
-      # rolling release is deleted and recreated each run, it would still look like it worked. Fail
-      # before anything is pushed instead.
-      - name: Assert source ref still declares SNAPSHOT versions
+      # The stamp step below would timestamp a bare `5.1.0` too, so this is not what makes the
+      # nightly safe — it is a tripwire: a source ref without a qualifier means a post-release bump
+      # was missed, and that should be seen rather than silently absorbed.
+      - name: Assert source ref still declares a pre-release product version
```
NipunaRanasinghe, line 165: Loosening this to accept a concrete extension version was what I was after in [my earlier comment](https://github.com/wso2/product-integrator/pull/2067#discussion_r3782946108) — dropping `wi.extension.version` from the assertion entirely goes further than that, and re-opens a failure this PR documents elsewhere.

`release.yml:13-16` in this same diff says the `<minor+2>` bump is load-bearing because `apply-version.sh` derives the pre-release extension line as `<minor-1>`, so it "only stays above the published stable while the declared version is the next *feature* release."

Concretely: a post-GA bump sets `integrator.version=5.2.0-SNAPSHOT` but leaves `wi.extension.version=1.2.0` (the version just released). The integrator check passes, nothing checks the extension, and `--force-extension` — now always passed at line 198 — derives `1.1.<yymmddHH>`, a pre-release *below* the published stable 1.2.0, and publishes it to the Marketplace. The rolling release being deleted and recreated means it still looks like it worked.

Keeping the loop over both keys and accepting `-SNAPSHOT` *or* an already-odd-minor concrete version for the extension covers both this and the original blocker.

.github/workflows/daily-build.yml
```diff
@@ -2,31 +2,35 @@ name: Daily Build
...
+# Pass builds/nightly-test / nightly-test for an isolated route that leaves production alone.
 #
 # Ad-hoc builds of your own branch are dev-build.yml; real releases are release.yml. Neither this
 # workflow nor those ever write to `main`.
 
 on:
   schedule:
-    # 12:00 PM IST = 06:30 AM UTC
-    - cron: '30 6 * * *'
+    - cron: '30 21 * * *'
```
NipunaRanasinghe, line 34: Non-blocking notes, batched here since several of them sit in parts of the diff the description doesn't cover.

Starting with this line: the cron moved from `30 6` to `30 21` (12:00 IST → 03:00 IST the next day) and the `# 12:00 PM IST = 06:30 AM UTC` comment was deleted rather than updated. Not mentioned in the description — either unintended, or intentional and it lost its rationale.

The rest:

- `resolve-nightly-versions.sh:95` — `select(.nightly == false and .qualified == false)` became `select(.nightly == false)`. The comment justifies it as staging-specific, but the script has no per-route switch and this is the same script the nightly runs when it falls back to `main`. Upstream cutting `v2201.13.0-rc1` means the next nightly off `main` ships that RC as the bundled runtime, where the old rule held at newest GA. Same for `icp.version`. Gating on the resolved route would keep both behaviours.
- `daily-build.yml:247` — `smoke_tests_required` is unconditionally `false` on `schedule`, and `gate_on_smoke_tests` defaults to `false` on dispatch, so nothing gates on smoke tests anywhere. With `run_smoke_tests: true` the suite still runs nightly and its only consumer is a `::warning::` in the called workflow's own job summary, which doesn't surface on the green parent run. Defaulting `gate_on_smoke_tests` to `true` would at least make a deliberate manual verification run gate.
- `daily-build.yml:115` — more than one `staging/*` branch exits 1 with no fallback. A `staging/5.1.0` left undeleted after GA while `staging/5.2.0` is cut stops nightly builds entirely, and on a `schedule` trigger the only signal is a red run in the Actions tab.
- `resolve-nightly-versions.sh:93` — `sort_by(.ts)` on the asset's `updated_at`. GitHub bumps that on any asset-row modification, not only a rolling republish, so a re-signed zip re-uploaded onto an older `v2201.12.x` moves that release to the front and the nightly pins backwards. `sort_by(.published)`, with the asset timestamp used only to disambiguate the rolling `nightly` entry, avoids it.
- `apply-version.sh:21-22` — the `--mode` usage line still says `release` "strips `-SNAPSHOT`". It now strips any `-[A-Za-z][A-Za-z0-9.]*` qualifier, which the shape table nine lines above already documents. The two now disagree.

Separately: the diff is well past what "What changed" lists — smoke-test `allow_failure` plumbing across `build.yml`/`smoke-test.yml`, this cron move, and the `resolve-nightly-versions.sh` candidate-selection rewrite. Worth updating the description before merge.

.github/workflows/daily-build.yml
samithkavishke: This is intended behavior

.github/workflows/daily-build.yml
samithkavishke: We will handle the validation logic separately.

.github/workflows/daily-build.yml
samithkavishke: Addressed

ci/build/resolve-nightly-versions.sh
samithkavishke: Addressed

NipunaRanasinghe:
Three small things left, none blocking:

1. The PR description still says "The production-branch safety guard now accepts `main` or any `staging/*` ref (previously `main` only)." The guard is deleted, which you've confirmed is intended — so that sentence now describes the opposite of the code. The header comment in `daily-build.yml` is accurate, it's only the PR body that isn't. Worth fixing before merge since it lands in the merge commit context.

2. `resolve-nightly-versions.sh:93` — `sort_by(.ts)` on the asset's `updated_at` wasn't touched and didn't get a reply, so I'm not sure whether it was considered and skipped or just missed. GitHub bumps `updated_at` on any asset-row modification, not only a rolling republish, so a corrected or re-signed zip re-uploaded onto an older `v2201.12.x` moves that release to the front and the next nightly pins backwards. `sort_by(.published)`, with the asset timestamp used only to disambiguate the rolling `nightly` entry, avoids it. Fine to leave if it's a deliberate trade.

3. `73d40f0` also dropped `# ballerina-custom-jre tags its releases without a leading "v" (e.g. 4.0.0).` above the JRE resolve call. That looks like collateral from removing the bullet above it — it was the only explanation for the empty fourth argument to `resolve_github_component`, which is otherwise unexplained at the call site.

Separately, the `workflow_dispatch` box in the test plan is still unchecked. That matters more now that the dispatch defaults are `builds/nightly`/`nightly`: an exercise run needs the test branch and tag typed explicitly, and it's the check that would confirm the staging resolution and the stamp step behave against `staging/5.1.0` end to end.

NipunaRanasinghe (approved):
Approved with non-blocking suggestions

samithkavishke:
> Three small things left, none blocking:
> 
> 1. The PR description still says "The production-branch safety guard now accepts `main` or any `staging/*` ref (previously `main` only)." The guard is deleted, which you've confirmed is intended — so that sentence now describes the opposite of the code. The header comment in `daily-build.yml` is accurate, it's only the PR body that isn't. Worth fixing before merge since it lands in the merge commit context.
> 2. `resolve-nightly-versions.sh:93` — `sort_by(.ts)` on the asset's `updated_at` wasn't touched and didn't get a reply, so I'm not sure whether it was considered and skipped or just missed. GitHub bumps `updated_at` on any asset-row modification, not only a rolling republish, so a corrected or re-signed zip re-uploaded onto an older `v2201.12.x` moves that release to the front and the next nightly pins backwards. `sort_by(.published)`, with the asset timestamp used only to disambiguate the rolling `nightly` entry, avoids it. Fine to leave if it's a deliberate trade.
> 3. `73d40f0` also dropped `# ballerina-custom-jre tags its releases without a leading "v" (e.g. 4.0.0).` above the JRE resolve call. That looks like collateral from removing the bullet above it — it was the only explanation for the empty fourth argument to `resolve_github_component`, which is otherwise unexplained at the call site.
> 
> Separately, the `workflow_dispatch` box in the test plan is still unchecked. That matters more now that the dispatch defaults are `builds/nightly`/`nightly`: an exercise run needs the test branch and tag typed explicitly, and it's the check that would confirm the staging resolution and the stamp step behave against `staging/5.1.0` end to end.

Will be addressed in separate PR

PR #2118, opened 2026-08-13
Add vsix publishing workflow

samithkavishke:
## Purpose
Currently product-integrator repository does not have a workflow to publish the vscode extension to the market place and there should be two options which is separate workflow to address this and as well as a step in the release process

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## New Features

* Added flexible VSIX release publishing.
* Releases can target the Visual Studio Code Marketplace, Open VSX, or both.
* Added support for selecting release tags and pre-release status.
* Marketplace publishing can be enabled after a successful build.
* Pre-release packages are correctly identified and published using the appropriate release mode.

### DIFF
--- a/.github/workflows/publish-vsix.yml
+++ b/.github/workflows/publish-vsix.yml
@@ -29,7 +29,9 @@ permissions:
 jobs:
   publish:
     uses: ./.github/workflows/reusable-publish-vsix.yml
-    secrets: inherit
+    secrets:
+      VSCE_TOKEN: ${{ secrets.VSCE_TOKEN }}
+      OPENVSX_TOKEN: ${{ secrets.OPENVSX_TOKEN }}
     with:
       tag: ${{ inputs.tag }}
       vscode: ${{ inputs.vscode }}
--- a/.github/workflows/publish-vsix.yml
+++ b/.github/workflows/publish-vsix.yml
@@ -12,11 +12,6 @@ on:
         description: 'Release tag to publish from, e.g. v5.1.0'
         required: true
         type: string
-      isPreRelease:
-        description: 'Publish as a pre-release'
-        required: true
-        type: boolean
-        default: false
       vscode:
         description: 'Publish to VS Code Marketplace'
         required: true
@@ -28,12 +23,14 @@ on:
         type: boolean
         default: true
 
+permissions:
+  contents: read
+
 jobs:
   publish:
     uses: ./.github/workflows/reusable-publish-vsix.yml
     secrets: inherit
     with:
       tag: ${{ inputs.tag }}
-      isPreRelease: ${{ inputs.isPreRelease }}
       vscode: ${{ inputs.vscode }}
       openVSX: ${{ inputs.openVSX }}
--- a/.github/workflows/publish-vsix.yml
+++ b/.github/workflows/publish-vsix.yml
@@ -1,0 +1,69 @@
+name: Publish WI Extension
+
+# Publishes the wso2-integrator VSIX already attached to a GitHub Release (created by release.yml)
+# to the VS Code Marketplace and/or Open VSX. This is a separate, manual step from release.yml:
+# the release always ships the vsix as a release asset; this workflow additionally pushes it out
+# to the marketplaces on demand.
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Release tag to publish from, e.g. v5.1.0'
+        required: true
+        type: string
+      isPreRelease:
+        description: 'Publish as a pre-release'
+        required: true
+        type: boolean
+        default: false
+      vscode:
+        description: 'Publish to VS Code Marketplace'
+        required: true
+        type: boolean
+        default: true
+      openVSX:
+        description: 'Publish to Open VSX marketplace'
+        required: true
+        type: boolean
+        default: true
+
+jobs:
+  publish:
+    name: Publish vsix to marketplaces
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Download WI extension VSIX from release
+        env:
+          TAG: ${{ inputs.tag }}
+        run: gh release download "${TAG}" --repo "${{ github.repository }}" --pattern 'wso2-integrator-*.vsix'
+
+      - name: Get vsix path
+        id: vsix
+        run: |
+          file=$(ls wso2-integrator-*.vsix)
+          echo "vsixName=${file}" >> "$GITHUB_OUTPUT"
+          if [ "${{ inputs.isPreRelease }}" == "true" ]; then
+            echo "releaseMode=--pre-release" >> "$GITHUB_OUTPUT"
+          else
+            echo "releaseMode= " >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Use Node.js 20.x
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20.x
+
+      - name: Install publishing CLIs
+        run: |
+          npm install -g @vscode/vsce ovsx
+
+      - name: Publish to VS Code Marketplace
+        if: ${{ inputs.vscode }}
+        run: vsce publish -p ${{ secrets.VSCE_TOKEN }} --packagePath ${{ steps.vsix.outputs.vsixName }} ${{ steps.vsix.outputs.releaseMode }}
+
+      - name: Publish to Open VSX marketplace
+        if: ${{ inputs.openVSX }}
+        run: ovsx publish -p ${{ secrets.OPENVSX_TOKEN }} --packagePath ${{ steps.vsix.outputs.vsixName }} ${{ steps.vsix.outputs.releaseMode }}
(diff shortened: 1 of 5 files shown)


.github/workflows/reusable-publish-vsix.yml
```diff
@@ -0,0 +1,68 @@
...
+        with:
+          node-version: 20.x
+
+      - name: Install publishing CLIs
+        run: |
+          npm install -g @vscode/vsce ovsx
+
+      - name: Publish to VS Code Marketplace
+        if: ${{ inputs.vscode }}
+        run: vsce publish -p ${{ secrets.VSCE_TOKEN }} --packagePath ${{ steps.vsix.outputs.vsixName }} ${{ steps.vsix.outputs.releaseMode }}
```
NipunaRanasinghe: **`vsce publish --pre-release` will always fail for this VSIX.**

`vsce` refuses `--pre-release` together with `--packagePath` unless the VSIX itself was *packaged* as pre-release — `@vscode/vsce` `out/publish.js` (lines 80-91) throws `Cannot use '--pre-release' flag with a package that was not packaged as pre-release`.

The WI VSIX is built by `wi/wi-extension` → `package-vsix.js`, which only adds `--pre-release` when the env var `isPreRelease=true`, and that variable is never set anywhere in `compile.yml` or `build.yml`. Since `release.yml`'s `is_prerelease` input defaults to `true`, `steps.vsix.outputs.releaseMode` is `--pre-release` on the default release path and this step aborts every time.

Fix by packaging the VSIX with `--pre-release` in the build, not by flagging it at publish time.

.github/workflows/reusable-publish-vsix.yml
```diff
@@ -0,0 +1,68 @@
...
+
+      - name: Publish to Open VSX marketplace
+        if: ${{ inputs.openVSX }}
+        run: ovsx publish -p ${{ secrets.OPENVSX_TOKEN }} --packagePath ${{ steps.vsix.outputs.vsixName }} ${{ steps.vsix.outputs.releaseMode }}
```
NipunaRanasinghe: **`ovsx --pre-release` is silently ignored for a prepackaged VSIX.**

`ovsx` `lib/publish.js` (line 46) logs `Ignoring option '--pre-release' for prepackaged extension.` and continues. So an alpha/beta build gets published to Open VSX as a normal (non-pre-release) version, and users on the stable channel receive it.

Combined with the `vsce` issue above, the `isPreRelease` input produces a hard failure on one registry and a wrong-but-successful publish on the other — neither matches what the input description promises.

NipunaRanasinghe: **Secrets and a release-asset filename are interpolated unquoted into `run:` bodies.**

Both publish steps expand `${{ secrets.* }}` and `${{ steps.vsix.outputs.vsixName }}` directly into the shell command. Two concrete failures:

1. If `VSCE_TOKEN` / `OPENVSX_TOKEN` is not configured in the repo, the expansion is empty, so `-p` swallows `--packagePath` and `wso2-integrator-1.3.x.vsix` is left as a stray positional arg — a confusing error instead of "missing secret".
2. `vsixName` comes from a GitHub release asset name, so an asset named e.g. `wso2-integrator-$(...).vsix` executes on the runner.

This also contradicts the convention the repo states explicitly at `release.yml:107-109` and `build.yml:1284-1285` (bind through `env`, never interpolate into a `run:` body) — the `TAG` step just above already follows it.

.github/workflows/reusable-publish-vsix.yml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Download WI extension VSIX from release
+        env:
+          TAG: ${{ inputs.tag }}
+        run: gh release download "${TAG}" --repo "${{ github.repository }}" --pattern 'wso2-integrator-*.vsix'
+
+      - name: Get vsix path
+        id: vsix
+        run: |
+          file=$(ls wso2-integrator-*.vsix)
+          echo "vsixName=${file}" >> "$GITHUB_OUTPUT"
```
NipunaRanasinghe: **`ls` glob with no `set -euo pipefail` and no single-match guard.**

If the release holds more than one matching VSIX, `$file` is multi-line and `echo "vsixName=${file}" >> "$GITHUB_OUTPUT"` writes a multi-line value with no heredoc delimiter — GitHub rejects it as an invalid output format. That is reachable today: the manual `publish-vsix.yml` accepts any tag, including the rolling `nightly` tag whose assets accumulate before `build.yml`'s delete-superseded step runs.

Add `set -euo pipefail` and fail explicitly when the match count is not exactly 1.

.github/workflows/release.yml
```diff
@@ -206,3 +211,18 @@ jobs:
       run_smoke_tests: ${{ inputs.run_smoke_tests }}
       ballerina_extension_source: ${{ inputs.ballerina_extension_source }}
       mi_extension_source: ${{ inputs.mi_extension_source }}
+
+  publish-marketplace:
+    name: Publish WI extension to marketplaces
+    needs: build
+    # The vsix only lands as a release asset when build_packed_installers produced the
+    # create-release-with-installers job's release (build.yml:1111); with it off, there is nothing
```
NipunaRanasinghe: **This comment names a job that does not exist.**

There is no `create-release-with-installers` job. The job that actually creates the release is `publish-github-release` (`build.yml:1095`); `build.yml`'s `create-release` job (line 1051) only uploads an artifact and never creates a release. The line reference is right, the job name is not — a reader chasing this gate will look for the wrong job.

.github/workflows/publish-vsix.yml
```diff
@@ -0,0 +1,39 @@
...
+        required: true
+        type: boolean
+        default: true
+      openVSX:
+        description: 'Publish to Open VSX marketplace'
+        required: true
+        type: boolean
+        default: true
+
+jobs:
```
NipunaRanasinghe, line 29: **No `permissions:` block, so `gh release download` depends on the org's default token scope.**

`reusable-publish-vsix.yml` declares no permissions either, so when invoked from this manual workflow the job runs with whatever the org's default `GITHUB_TOKEN` permissions are. If that default is restricted/none, `gh release download` fails with a 404 that reads like a missing tag. The `release.yml` path only works because that caller declares `contents: write` (`release.yml:89-91`).

Declare `permissions: contents: read` on the reusable workflow.

NipunaRanasinghe (changes requested):

.github/workflows/reusable-publish-vsix.yml
samithkavishke: Addressed

.github/workflows/reusable-publish-vsix.yml
samithkavishke: Addressed

samithkavishke: Addressed

.github/workflows/reusable-publish-vsix.yml
samithkavishke: Addressed

.github/workflows/release.yml
samithkavishke: Addressed

.github/workflows/publish-vsix.yml
samithkavishke: Addressed

.github/workflows/publish-vsix.yml
```diff
@@ -0,0 +1,36 @@
...
+        type: boolean
+        default: true
+
+permissions:
+  contents: read
+
+jobs:
+  publish:
+    uses: ./.github/workflows/reusable-publish-vsix.yml
+    secrets: inherit
```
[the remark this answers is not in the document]
samithkavishke: Addressed

.github/workflows/reusable-publish-vsix.yml
```diff
@@ -0,0 +1,99 @@
...
+      # same Microsoft.VisualStudio.Code.PreRelease manifest property @vscode/vsce's own `publish`
+      # command checks before it will accept `--pre-release --packagePath`.
+      - name: Detect pre-release from vsix manifest
+        id: prerelease
+        shell: bash
+        env:
+          VSIX_NAME: ${{ steps.vsix.outputs.vsixName }}
+        run: |
+          set -euo pipefail
+          if unzip -p "${VSIX_NAME}" extension.vsixmanifest | grep -q 'Id="Microsoft.VisualStudio.Code.PreRelease"'; then
```
NipunaRanasinghe, line 70: Nit, non-blocking: this pipeline runs under `set -o pipefail`. If `grep -q` matches, it exits immediately and closes the pipe; `unzip` can then take SIGPIPE, making the pipeline status non-zero, so the `if` takes the `else` branch and a pre-release vsix publishes to the stable channel.

In practice a vsixmanifest is a few KB — well under the 64KB pipe buffer — so `unzip` writes it all and exits before `grep` reads, and this won't fire. But the failure is silent and lands in the wrong channel, so it's cheap to remove the race:

```bash
unzip -p "${VSIX_NAME}" extension.vsixmanifest > manifest.xml
if grep -q 'Id="Microsoft.VisualStudio.Code.PreRelease"' manifest.xml; then
```

NipunaRanasinghe (dismissed):
Approving. All six points from my earlier review are addressed in the latest commits.

One non-blocking nit inline on the manifest-detection step. both secrets are `required: false`, so an unconfigured token becomes `-p ""` and surfaces as an auth failure rather than "secret not set".

NipunaRanasinghe (approved):

PR #2178, opened 2026-08-19
Change the release and daily build pipelines to run on GitHub-hosted runners by default.

samithkavishke:
## Goals
$Subject

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary by CodeRabbit

* **Build and Release Improvements**
  * Builds and releases now use GitHub-hosted runners by default, with CodeBuild still available as an option.
  * Platform artifacts continue to be provided through GitHub Actions.
  * Increased compile job timeout to improve reliability for longer builds.
  * Added disk-space cleanup and usage reporting to help prevent build failures.
  * Improved package-building checks for RPM-based environments.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -579,20 +579,6 @@ jobs:
           ./build.sh ballerina.zip $BALLERINA_VERSION VSCode-darwin-${{ matrix.arch }}.zip icp.zip jre.zip $INTEGRATOR_VERSION ${{ matrix.arch }}
           cd ${ROOT_DIR}
 
-      - name: Login to AWS
-        uses: aws-actions/configure-aws-credentials@v5
-        with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          aws-region: us-east-1
-      
-      - name: Upload macOS artifacts to S3
-        if: ${{ inputs.build_packed_installers == true }}
-        working-directory: .
-        run: |
-          S3_PATH="s3://${{ secrets.AWS_S3_BUCKET }}/wi/${{ github.run_id }}/"
-          aws s3 cp installers/mac/ "${S3_PATH}" --recursive --exclude "*" --include "*.dmg"
-      
       - name: Upload macOS DMG artifacts
         if: ${{ inputs.build_packed_installers == true }}
         uses: actions/upload-artifact@v4
@@ -996,21 +982,6 @@ jobs:
           Invoke-WebRequest -Uri $ICP_URL -OutFile "icp.zip"
           Invoke-WebRequest -Uri $JRE_URL -OutFile "jre.zip"
           .\build.bat .\ballerina.zip $BALLERINA_VERSION .\VSCode-win32-x64.zip .\icp.zip .\jre.zip $INTEGRATOR_VERSION
-
-      - name: Login to AWS
-        uses: aws-actions/configure-aws-credentials@v5
-        with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          aws-region: us-east-1
-
-      - name: Upload Windows artifacts to S3
-        if: ${{ inputs.build_packed_installers == true }}
-        shell: bash
-        working-directory: .
-        run: |
-          S3_PATH="s3://${{ secrets.AWS_S3_BUCKET }}/wi/${{ github.run_id }}/"
-          aws s3 cp installers/windows/WixPackage/bin/x64/Release/en-US/ "${S3_PATH}" --recursive --exclude "*" --include "*.msi"
 
       - name: Upload Windows artifacts
         if: ${{ inputs.build_packed_installers == true }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,7 +10,7 @@ on:
       runner:
         description: 'Select runner type'
         required: false
-        default: 'codebuild'
+        default: 'ubuntu-latest'
         type: string
       build_linux:
         description: 'Build for Linux'
@@ -90,9 +90,6 @@ on:
 permissions:
   id-token: write
   contents: write     
-
-env:
-  awsRunnerId: codebuild-wso2_product-integrator-${{ github.run_id }}-${{ github.run_attempt }}  
 
 jobs:
   # Settle the product and extension versions ONCE, up front, and hand the resulting files to every
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -804,20 +804,3 @@ jobs:
           name: wi-extension-vsix
           path: wi/wi-extension/wso2-integrator-*.vsix
           if-no-files-found: error
-
-      
-      - name: Login to AWS
-        if: ${{ inputs.build_packed_installers == true && github.event_name != 'pull_request' }}
-        uses: aws-actions/configure-aws-credentials@v5
-        with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          aws-region: us-east-1
-      
-      - name: Upload linux artifacts to S3
-        if: ${{ inputs.build_packed_installers == true && github.event_name != 'pull_request' }}
-        working-directory: .
-        run: |
-          S3_PATH="s3://${{ secrets.AWS_S3_BUCKET }}/wi/${{ github.run_id }}/"
-          aws s3 cp installers/linux-deb/ "${S3_PATH}" --recursive --exclude "*" --include "*.deb"
-          aws s3 cp installers/linux-rpm/ "${S3_PATH}" --recursive --exclude "*" --include "*.rpm"
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -172,15 +172,24 @@ jobs:
   compile:
     needs: resolve-versions
     name: Build repo
-    # 90 gives a fully cold-cache build room to finish and reach the post-step that SAVES the
-    # caches; warm runs finish well under this. 45 was too tight for a cold build, and a
-    # timeout there cancels the job before caches are saved -> every run stays cold (vicious cycle).
-    timeout-minutes: 90
+    timeout-minutes: 120
     runs-on: ${{ inputs.runOnAWS && inputs.awsRunnerId || 'ubuntu-latest' }}
     defaults:
       run:
         working-directory: lib/vscode
     steps:
+      - name: Free disk space
+        if: ${{ !inputs.runOnAWS }}
+        uses: jlumbroso/free-disk-space@v1.3.1
+        with:
+          tool-cache: false
+          android: true
+          dotnet: true
+          haskell: true
+          large-packages: false
+          docker-images: true
+          swap-storage: false
+
       - name: Checkout
         uses: actions/checkout@v4
         with:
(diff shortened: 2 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -44,11 +44,11 @@ on:
       runner:
         description: 'Select runner type'
         required: true
-        default: 'codebuild'
+        default: 'ubuntu-latest'
```
NipunaRanasinghe, line 47: This makes `ubuntu-latest` the default for real releases, but nothing in this PR has exercised that path.

`pr-ci.yml` only triggers on `main` and `5.0.x`, and this targets `staging/5.1.0`, so no build check ran here at all. On top of that, `pr-ci` calls `compile.yml` with `build_packed_installers: false`, so the installer part of the compile job has never actually run on `ubuntu-latest` in CI. That is exactly what this PR is switching on by default.

Could you link a green `dev-build` run with `runner=ubuntu-latest`, all three platforms, and installers enabled? Hard to sign off on the release default without one.

.github/workflows/build.yml
```diff
@@ -582,20 +579,6 @@ jobs:
           ./build.sh ballerina.zip $BALLERINA_VERSION VSCode-darwin-${{ matrix.arch }}.zip icp.zip jre.zip $INTEGRATOR_VERSION ${{ matrix.arch }}
           cd ${ROOT_DIR}
 
-      - name: Login to AWS
```
NipunaRanasinghe, line 585: Dropping the S3 uploads is a separate change from the runner switch, and neither the title nor the description mentions it. Worth calling out explicitly.

I checked the in-repo side and it looks safe: `s3://` appears only in these three spots (mac here, Windows below, Linux in `compile.yml`), and `resign-and-rehash.yml` pulls installers from GitHub Releases, not S3. So nothing in the repo breaks.

The open question is people outside the repo. Does anyone pull installers from `s3://.../wi/<run_id>/` — QA, release engineering, anyone with a bookmarked path? If so this breaks quietly, with no error anywhere.

If the removal stands, `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` and `AWS_S3_BUCKET` become unused and should be revoked. Also worth saying: getting long-lived static AWS keys out of a job that does signing is a real improvement.

.github/workflows/compile.yml
```diff
@@ -172,15 +172,24 @@ jobs:
...
-    timeout-minutes: 90
+    timeout-minutes: 120
     runs-on: ${{ inputs.runOnAWS && inputs.awsRunnerId || 'ubuntu-latest' }}
     defaults:
       run:
         working-directory: lib/vscode
     steps:
+      - name: Free disk space
+        if: ${{ !inputs.runOnAWS }}
+        uses: jlumbroso/free-disk-space@v1.3.1
```
NipunaRanasinghe: Please pin this to a commit SHA rather than the `v1.3.1` tag. Tags are movable, and this job runs with `secrets: inherit` and `id-token: write` for cosign signing, with the action running before checkout.

The repo already does this elsewhere — `resign-and-rehash.yml` uses `sigstore/cosign-installer@59acb6260d9c0ba8f4a2f9d9b48431a222b68e20`.

Same applies to `awalsh128/cache-apt-pkgs-action@latest` a bit further down. That one predates this PR, but `@latest` in a signing job is worse and it is cheap to fix while you are here.

.github/workflows/compile.yml
```diff
@@ -172,15 +172,24 @@ jobs:
   compile:
     needs: resolve-versions
     name: Build repo
-    # 90 gives a fully cold-cache build room to finish and reach the post-step that SAVES the
-    # caches; warm runs finish well under this. 45 was too tight for a cold build, and a
-    # timeout there cancels the job before caches are saved -> every run stays cold (vicious cycle).
-    timeout-minutes: 90
+    timeout-minutes: 120
```
NipunaRanasinghe, line 179: The comment that got deleted here carried a warning that still applies at 120: if the job times out, it is cancelled before the post-step that saves the caches, so the next run starts cold too and the timeout keeps repeating. That mechanism is unchanged by the bump. Please put an updated version of the comment back — the rest of these workflows document this kind of decision consistently.

For calibration: the codebuild compile job took about 43 minutes (run 32189289556). `ubuntu-latest` without installers runs 30-39 minutes in recent `pr-ci` runs. Add installers plus 2-4 minutes for the disk cleanup step and 120 looks reasonable, but nobody has measured the real thing yet.

.github/workflows/compile.yml
```diff
@@ -172,15 +172,24 @@ jobs:
...
     runs-on: ${{ inputs.runOnAWS && inputs.awsRunnerId || 'ubuntu-latest' }}
```
NipunaRanasinghe, line 180: On the CodeRabbit comment here claiming the `awsRunnerId` default expression is never evaluated so codebuild gets an invalid runner label — that is not right. Daily builds run on codebuild today and pass; run 32189289556 from Aug 18 shows the compile job on runner `e8da83d3-...`. No change needed.

The real version of that worry is different: once codebuild stops being the default, that path barely runs, and if it does break nobody will notice for a while. Worth deciding whether you keep it as a supported fallback or drop it.

Separately, removing the `awsRunnerId` env block from `build.yml` is fine. I checked the base branch and it was referenced nowhere; `compile.yml` computes its own default.

.github/workflows/compile.yml
```diff
@@ -308,16 +317,16 @@ jobs:
       - name: Cache apt packages
         uses: awalsh128/cache-apt-pkgs-action@latest
         with:
-          packages: build-essential g++ libx11-dev libxkbfile-dev libsecret-1-dev libkrb5-dev python-is-python3 quilt
+          packages: build-essential g++ libx11-dev libxkbfile-dev libsecret-1-dev libkrb5-dev python-is-python3 quilt rpm
```
NipunaRanasinghe, line 328: This one is load-bearing, not incidental cleanup — worth a note in the description. The codebuild image had `rpmbuild`; `ubuntu-latest` does not. `pr-ci` never caught it because it runs with `build_packed_installers: false`.

The `version: 1.1 -> 1.2` bump is required for the new package to actually be installed rather than served from the old cache, and adding `rpmbuild` to the verify loop below means it fails loudly instead of producing a missing RPM. Both correct.

.github/workflows/compile.yml
```diff
@@ -617,6 +626,11 @@ jobs:
           wget -O jre.zip "$JRE_URL"
           ./build.sh ballerina.zip $BALLERINA_VERSION vscode-linux-x64.tar.gz icp.zip jre.zip $INTEGRATOR_VERSION
           cd ${ROOT_DIR}
+      # always(): the point of this probe is the run where the installer tail ran out of disk.
+      - name: Report disk usage after packaging
```
NipunaRanasinghe, line 639: One `df -h` costs nothing, so keeping it is fine. But the comment frames it as a probe for one specific past failure, which reads like debug code that was meant to come out. Either reword it as a permanent diagnostic, or drop the step once you have a green installer run on `ubuntu-latest`.

The `free-disk-space` settings above look sensible by the way: roughly 19 GB reclaimed, tool cache kept since Node and Java need it, `large-packages` skipped since it is slow for what it returns.

.github/workflows/dev-build.yml
```diff
@@ -4,7 +4,7 @@ name: Development Build
 #
 # Versions are timestamped for this run only: a `-SNAPSHOT` branch builds as 5.1.0-<yyyymmddHHmm>
 # with the extension at 1.1.<yymmddHH>. Nothing is committed to any branch and nothing is published
-# to GitHub Releases; you get the workflow artifacts and the usual S3 upload.
+# to GitHub Releases; you get the workflow artifacts.
```
NipunaRanasinghe, line 7: This is the case I would double-check before merging. Dev builds publish nothing to Releases, so after this PR the only way to get an installer out of one is the Actions artifact: wrapped in a zip, needs a GitHub login, gone after 90 days. The S3 path had none of those constraints.

If anyone is currently grabbing dev-build DMGs or MSIs from the S3 bucket, this changes their workflow without telling them. Worth checking with whoever uses dev builds.

.github/workflows/daily-build.yml
```diff
@@ -239,7 +239,7 @@ jobs:
       # build.yml replaces only the configured rolling release/tag, whichever that is.
       publish_tag: ${{ needs.prepare-nightly.outputs.publish_tag }}
       is_prerelease: true
-      runner: codebuild
+      runner: ubuntu-latest
```
NipunaRanasinghe: This is hardcoded and there is no dispatch input to override it, so if GitHub-hosted runners turn out flaky the only way to put the nightly back on codebuild is a code change and a merge. That is a slow fix for something that fails at 03:00 IST.

Suggest adding a `runner` choice input to the `workflow_dispatch` block, defaulting to `ubuntu-latest`, and passing it through here. Same shape as `release.yml`.

.github/workflows/build.yml
```diff
@@ -10,7 +10,7 @@ on:
       runner:
         description: 'Select runner type'
         required: false
-        default: 'codebuild'
+        default: 'ubuntu-latest'
```
NipunaRanasinghe, line 13: Minor: this default never actually applies. `build.yml` is `workflow_call` only, and both callers pass `runner` explicitly — `release.yml` forwards `inputs.runner`, `daily-build.yml` hardcodes it. Harmless to change for consistency, just noting that the real switches are in `dev-build.yml`, `release.yml` and `daily-build.yml`.

.github/workflows/compile.yml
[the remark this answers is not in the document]
samithkavishke: Not applicable — `github.run_id` does resolve in a `workflow_call` input default.

Daily build 32189289556 ran with `runner: codebuild` and the compile job's label
resolved to `codebuild-wso2_product-integrator-32189289556-1` (run id + attempt
interpolated), on a self-hosted runner, successfully.

`awsRunnerId` has never been passed by a caller — `compile.yml`'s own default has
always produced that label. github/docs#29812 concerns the `inputs` context, not
`github`; the `github` context is available in this field.

`compile.yml:10` is also unchanged by this PR. The `env: awsRunnerId` block removed
from build.yml was dead code — top-level `env` can't reach a reusable workflow's
`with:`.

NipunaRanasinghe (changes requested):

.github/workflows/release.yml
samithkavishke: Pending — will trigger a dev-build with `runner=ubuntu-latest`, all three platforms and installers enabled, and link the green run here before merge.

.github/workflows/build.yml
samithkavishke: Addressed — checked, nothing outside the repo consumes `s3://.../wi/<run_id>/`; updating the title/description to call out the removal, and the three AWS secrets will be revoked after merge.

.github/workflows/compile.yml
samithkavishke: Addressed — pinned `free-disk-space` to `54081f138730dfa15788a46383842cd2f914a1be` (v1.3.1) and `cache-apt-pkgs-action` to `553a35bb8ebd9fcabcb1c9451aa4c98e1b4ca8a9` (v1.6.3); no unpinned third-party actions left in this job.

.github/workflows/compile.yml
samithkavishke: Addressed — restored the comment with the self-perpetuating-timeout rationale, updated for the 90 -> 120 bump.

.github/workflows/compile.yml
samithkavishke: Not applicable — agreed, no change needed; keeping codebuild as a selectable fallback, and daily-build now has a `runner` input so it can be exercised.

.github/workflows/compile.yml
samithkavishke: Addressed — will call this out explicitly in the description as a load-bearing fix rather than cleanup.

.github/workflows/compile.yml
samithkavishke: Addressed — reworded as a permanent diagnostic instead of a probe for one past failure.

.github/workflows/dev-build.yml
samithkavishke: Addressed — checked, nobody outside the repo pulls dev-build installers from the bucket, so no downstream workflow changes; artifacts are sufficient.

.github/workflows/daily-build.yml
samithkavishke: Addressed — added a `runner` choice input defaulting to `ubuntu-latest` and passing through as `${{ inputs.runner || 'ubuntu-latest' }}`, so the scheduled nightly stays on ubuntu-latest while a manual dispatch can pick codebuild.

.github/workflows/build.yml
samithkavishke: Not applicable — consistency-only, agreed; all three callers pass `runner` explicitly, so this default is never the deciding value.

NipunaRanasinghe (approved):
LGTM

PR #2213, opened 2026-08-20
Fix washed-out SMB and FTP icons in the project explorer

niveathika:
## Purpose

The SMB and FTP entries in the Entry Points tree rendered faint and undersized next to their neighbours (Automation, Local Files, HTTP Service).

The cause turned out to be **sub-pixel stroke weight, not glyph size**. At the 16px tree size SMB's strokes measured 0.75px and FTP's 0.88px, so every line antialiased to grey instead of resolving to a black pixel. Measured at 16px, `darkest pixel` is the value each icon's blackest pixel reaches:

| icon | stroke @256 | → px @16 | darkest pixel | solid black px |
|---|---|---|---|---|
| `bi-file` (renders correctly) | 16 | 1.00 | 0 | 86 |
| `bi-globe` (renders correctly) | 16 | 1.00 | 0 | 59 |
| `bi-ftp` | 14 | 0.88 | 31 | 48 |
| `bi-smb` | 12 | 0.75 | **49** | 12 |

SMB never produced a single fully-black pixel. Every icon that looks right hits a true 0.

## Approach

Reduce the `viewBox` padding on both so the glyph scales up. This helps twice over — it thickens the stroke *and* aligns it to the pixel grid — and it measured better than bumping `stroke-width` alone (`stroke-width="3"` on SMB with padding kept only reached darkest 27, vs 17 for padding removal).

FTP's `stroke-width` is on the root `<svg>`, so it applies to the "FTP" lettering as well as the folder outline; scaling the viewBox alone thickened the type. Scaling `stroke-width` by the same factor (`0.8 → 0.74 = 0.8 × 31.92/34.72`) keeps the lettering at its original weight.

| file | change |
|---|---|
| `light/dark-bi-smb.svg` | `viewBox` `-8.77 -8.77 217.34 217.34` → `4.82 4.82 190.17 190.17` |
| `light/dark-bi-ftp.svg` | `viewBox` `-2.36 -2.36 34.72 34.72` → `-0.96 -0.96 31.92 31.92`, `stroke-width` `0.8` → `0.74` |

Path data is untouched. Licence headers unchanged.

## Results

| icon | box @256 | darkest @16 | solid black px | grey px |
|---|---|---|---|---|
| `bi-smb` | 198×224 → 226×256 | 49 → **17** | 12 → **52** | 75 → 47 |
| `bi-ftp` | 228×228 → 246×246 | 31 → **22** | 48 → **56** | 35 → 36 |

FTP's stroke stays at 14 (unchanged from before), so only its size and pixel alignment improve.

## Notes

- SMB's stroke is now 0.86px at 16px — better, but still under one device pixel, so its darkest pixel is 17/255 rather than a true 0. Closing that last gap would need the stroke thickened relative to the glyph, which changes the drawn weight. 17/255 reads as black, so it is left alone here.
- The "SMB"/"FTP" lettering remains illegible at 16px. That is inherent to three letters in a 16px tile and is unchanged by this PR.
- **Not addressed here:** `bi-azure-files` is also off-style — it is a potrace-traced bitmap (`viewBox="0 0 1254 1254"` under `transform="scale(0.1,-0.1)"`, one path, 420 coordinates), rendering as a solid filled folder with stray trace artifacts at 44.4% ink against the family's 21–35%. It has no padding to remove and already hits a true black, so no viewBox change helps; it needs redrawn artwork. Worth a separate issue.

## Testing

Rendered every `light-bi-*.svg` in `wi/wi-extension/assets/` with `rsvg-convert` at 16px and 256px, and measured ink bounding box, modal stroke run-length, darkest pixel, and solid-vs-grey pixel counts. Verified the Entry Points tree order from the reported screenshot reproduces before the change and is corrected after.


samithkavishke (approved):
LGTM

PR #2245, opened 2026-08-28
feat(macos): Developer ID signing and notarization for the macOS build

gigara:
The macOS build produced an app and DMG that Gatekeeper blocks on any machine other than the one that built them: nothing imported a Developer ID certificate, and nothing submitted the result to Apple. This adds signing and notarization to the existing build.

## What it does

- **Imports both certificates into a temporary keychain**: Developer ID Application (`codesign`, for the app bundle and DMG) and Developer ID Installer (`productbuild` — an unsigned `.pkg` cannot be notarized), then derives the identity names from the imported certs so no separate identity secret is needed.
- **Sanity-checks the decoded PKCS#12 before importing** (a DER SEQUENCE starts with `0x30`), which separates "the secret was mangled in transit" from "the password is wrong" — `security import` reports both as the same MAC verification failure.
- **Signs hardened-runtime with entitlements, deepest-first**: loose Mach-O and bundled component executables, then nested bundles, then the top-level app, so a nested binary is never re-signed after the bundle enclosing it. The DMG container itself is also signed.
- **Notarizes and staples** the DMG and PKG via `notarytool`, failing loudly if the artifacts are missing rather than skipping silently.

The entitlements are a superset of VS Code's own because one plist deep-signs the bundled JVM too (`allow-jit`, `allow-unsigned-executable-memory`, `disable-library-validation`, `allow-dyld-environment-variables` — the set notarized JDK distributions ship with). Debugger and sandbox-only entitlements are deliberately not included.

## Secrets

`MACOS_CERTIFICATE(_PASSWORD)`, `MACOS_INSTALLER_CERTIFICATE(_PASSWORD)`, `KEYCHAIN_PASSWORD`, `NOTARY_APPLE_ID`, `NOTARY_APP_PASSWORD`, `NOTARY_TEAM_ID`. Every signing step is skipped when they are absent, so a fork without them still builds an ad-hoc-signed app rather than failing.

## Review

Reviewed pre-PR by CodeRabbit (two rounds) plus a scoped manual review; all findings fixed — including a working-directory bug that made the notarize step silently succeed while notarizing nothing, which is why the fail-if-missing check exists. The verification round returned zero findings.

Part 1 of 2: the component update mechanism PR builds on this (Squirrel.Mac requires a validly signed app to swap).

## Summary by CodeRabbit

* **New Features**
  * macOS installers can now be signed with Developer ID certificates.
  * macOS DMG and PKG installers can be notarized and stapled for improved distribution readiness.
  * Applications use the configured bundled or external Java runtime when available.

* **Bug Fixes**
  * Improved DMG cleanup reliability with retries, diagnostics, and forced unmount fallback.
  * Added required entitlements for signed application runtime components.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -547,6 +547,76 @@ jobs:
           mkdir -p ../artifacts
           (cd .. && zip -qry "artifacts/VSCode-darwin-${{ matrix.arch }}.zip" VSCode-darwin-${{ matrix.arch }})
           cp ../artifacts/VSCode-darwin-${{ matrix.arch }}.zip ../../installers/mac/
+      # Imports the Developer ID Application cert (codesign: app bundles + DMG) and the
+      # Developer ID Installer cert (productbuild: the .pkg — an unsigned pkg cannot be
+      # notarized) into a temp keychain, then derives the identity names from the imported
+      # certs so no separate identity secret is needed.
+      - name: Import signing certificates (macOS)
+        if: ${{ inputs.build_packed_installers == true }}
+        env:
+          MACOS_CERTIFICATE: ${{ secrets.MACOS_CERTIFICATE }}
+          MACOS_CERTIFICATE_PASSWORD: ${{ secrets.MACOS_CERTIFICATE_PASSWORD }}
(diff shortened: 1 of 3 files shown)


gigara:
Recreating from an in-repo branch so the pair forms a real stack (auto-retarget on merge). Replacement linked shortly.

PR #2246, opened 2026-08-28
feat(update): component-wise update mechanism

gigara:
> **Stacked on #2245**: this PR's base is that PR's branch, so the diff below is the update mechanism alone. Merge #2245 first, then retarget this PR to `main` (one click) and merge.

Lets a shipped Integrator update its bundled components — the Ballerina runtime, the JRE, ICP, and the six extensions — and the app itself, without waiting for a full product release.

## Shape

The client posts its inventory to `POST /api/v1/updates` and the **server decides** what to offer, composing each response from one cosign-signed source document per release. Rollout, entitlement and kill-switch decisions live server-side, changeable without shipping a client. A per-channel `index.json` maps a client-version selector (`5.1.x`, an exact version, a range) to the document that serves it — first match wins — so a 5.1 maintenance line and a 5.2 line coexist without either touching the other's document.

## Trust

Two things are signed, verified by different parties: the **source document** (checked by the update server) and a per-artifact **statement** binding `{id, version, sha256, sizeBytes, requires}` (checked by the client against a key pinned in `product.json` at build time). Signing the statement rather than the bytes is what stops a manipulated document offering an old, genuinely signed artifact under a new version label — and it is why the server can compose responses per client without ever holding the signing key. CI compares the pinned key against the signing key before every publish, since a mismatch would otherwise surface only as every client rejecting every artifact, after release.

On the machine: runtime components are staged and swapped at the next launch — the only moment nothing runs from the directories being replaced — and an activated runtime that keeps the app from booting is rolled back automatically after two failed starts. The app applies via Squirrel.Mac (macOS) or an in-place user-context MSI (Windows). No privilege elevation anywhere.

## Publishing

Opt-in by design: `build.yml` takes `publish_update_source` defaulting to **false**; only the Release workflow sets it (default true, switchable off) — so a nightly cannot claim a release line by accident. A separate `publish-components.yml` ships a component fix with no app build, tag or installers, in minutes. Both publishers share one index-upsert script (unreachable-entry and sequence-regression guards) and a per-channel concurrency group; sequences are UTC clock reads so the publishers never coordinate numbering. `promote-channel.yml` copies already-signed documents down the insider→beta→stable ladder, per line.

## Verified

On real hardware against a deployed server: discovery, download, statement verification, staging, activation and `${COMPONENT:*}` resolution against a live Ballerina process; three forged artifacts and a tampered document all refused; channel selection and insider→beta promotion; a component-only publish delivering an extension update with no app release. Windows in-place MSI apply and macOS Squirrel were verified under the earlier flow. Not yet exercised: Linux, and Windows against the server-decided flow.

## Review

Multi-round pre-PR review (three scoped manual reviewers + CodeRabbit, two rounds): 7 critical and ~15 warning findings fixed — among them server-supplied id/version path validation before any filesystem use, refusing HTTP redirects (which would have leaked the Authorization header and escaped the host allowlist), fail-closed verification when a required pinned key is unusable, and promotion aborting on transient read failures instead of rebuilding the target index. The verification round returned clean; remaining known risks (signing-key rotation → KMS, trust-on-publish for upstream artifacts, unenforced statement `sizeBytes`) are tracked in the threat model's risk registry.

## Secrets

`WSO2_UPDATE_URL`, `WSO2_UPDATE_PUBLIC_KEY`, `WSO2_UPDATE_ARTIFACTS_URL`, `AWS_UPDATE_S3_BUCKET`, `COSIGN_*`. All optional: without them the updater ships disabled rather than broken. The update server is deployed separately and is not in this repository.


gigara:
Recreating from an in-repo branch so the pair forms a real stack. Replacement linked shortly.

PR #2254, opened 2026-08-31
Add markCodeServerDeployed to update draft status in cloud editor

lckariyawasam:
Currently a draft status is stored in the cloud editor to show the user that they have an undeployed draft. This needs to be removed once the draft component is deployed. The function to do that is already registered https://github.com/wso2-enterprise/integration-web-editor/pull/20. However, this needs to be called from the WI extension once the component is deployed

Execution of the command is flagged to run only on cloud editor env so it will not be called on desktop app.

Resolves https://github.com/wso2-enterprise/integration-engineering/issues/2264

### DIFF
--- a/wi/wi-extension/src/cloud/cmds/create-component-cmd.ts
+++ b/wi/wi-extension/src/cloud/cmds/create-component-cmd.ts
@@ -325,6 +325,7 @@ export const submitCreateComponentHandler = async ({ createParams, org, project,
 
 	if (result.created.length > 0) {
 		clearCodeServerLocalStorage();
+		markCodeServerDeployed();
 		const projectCache = dataCacheStore.getState().getProjects(org?.handle);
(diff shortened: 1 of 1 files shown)


wi/wi-extension/src/cloud/cmds/create-component-cmd.ts
```diff
@@ -325,6 +325,7 @@ export const submitCreateComponentHandler = async ({ createParams, org, project,
 
 	if (result.created.length > 0) {
 		clearCodeServerLocalStorage();
+		markCodeServerDeployed();
```
pcnfernando, line 328: Shouldn't this go inside `clearCodeServerLocalStorage`?

pcnfernando: clearCodeServerLocalStorage() needs to be called independently

pcnfernando (approved):
LGTM

NipunaRanasinghe (approved):

PR #2269, opened 2026-09-01
[Cloud Edior] Add fallback option to list MCP Proxy components

pcnfernando:
## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved component creation when source component details are unavailable from workspace data.
  * Added fallback retrieval of component subtype information to improve MCP proxy repository attachments.
  * Improved attachment handling when a component name is not otherwise available.
  * Added graceful error handling for missing component metadata.

### DIFF
--- a/wi/wi-extension/src/cloud/cmds/create-component-cmd.ts
+++ b/wi/wi-extension/src/cloud/cmds/create-component-cmd.ts
@@ -42,7 +42,7 @@ import { dataCacheStore } from "../stores/data-cache-store";
 import { isSamePath, isSubpath } from "../../utils/pathUtils";
 import { getUserInfoForCmd, isRpcActive, selectOrg, selectProjectWithCreateNew, setExtensionName } from "./cmd-utils";
 import { updateContextFile } from "./create-directory-context-cmd";
-import { attachMCPProxyRepositoryToExistingTrack, isMcpProxyFromExistingApi } from "../graphql/cp-graphql-client";
+import { attachMCPProxyRepositoryToExistingTrack, fetchComponentSubType, isMcpProxyFromExistingApi } from "../graphql/cp-graphql-client";
 import { WICloudSubmitComponentsReq, WICloudSubmitComponentsResp } from "@wso2/wi-core";
 import { openCloudFormWebview } from "../../ws-managers/cloud/ws-manager";
 import { ProjectType, StateMachine, stateService } from "../../stateMachine";
@@ -260,14 +260,48 @@ export const submitCreateComponentHandler = async ({ createParams, org, project,
 	const dotGit = await newGit?.getRepositoryDotGit(workspaceFsPath);
 	const repo = newGit.open(gitRoot, dotGit);
 
-	const workspaceCompId: string | null | undefined = ext.context.workspaceState.get("SOURCE_COMPONENT_ID");
+	// The cloud editor injects the source component through the environment, while the clone and
+	// prebuilt flows persist it in workspace state, so both have to be consulted.
+	const stateCompId: string | null | undefined = ext.context.workspaceState.get("SOURCE_COMPONENT_ID");
+	const envCompId: string | undefined = process.env.SOURCE_COMPONENT_ID;
+	const workspaceCompId: string | null | undefined = stateCompId || envCompId;
+	// Logged unconditionally: an absent line here means this build predates the source-component
+	// handling, which is otherwise indistinguishable from the id simply not being set.
+	ext.log(
+		`Resolving source component: workspaceState='${stateCompId ?? ""}', env='${envCompId ?? ""}', ` +
+			`resolved='${workspaceCompId ?? ""}', cloudEditor=${ext.isDevantCloudEditor}, cloudEnv=${ext.cloudEnv}`,
+	);
 	if (workspaceCompId) {
 		const component = dataCacheStore.getState().getComponents(org.handle, project.handler)?.find(comp => comp.metadata?.id === workspaceCompId);
 		if (component?.metadata?.isPrebuilt) {
 			// if its pre-built integration, we need to update the existing component with new repo details instead of creating a new component.
 			return await handlePrebuiltComponentUpdate(workspaceCompId, component, org, project, createParams[0], workspaceFsPath, gitRoot!);
 		}
-		if (isMcpProxyFromExistingApi(component)) {
+
+		// Prefer the sub-type from the component list. Only the CLI's `withSystemComponents` query
+		// selects `componentSubType`, so if `component/getList` used the plain query it arrives empty —
+		// in that case ask the control plane directly. A lookup failure must not block a normal
+		// deployment, so fall through to creating a component.
+		let componentSubType: string | undefined = component?.spec?.subType || undefined;
+		let subTypeSource = "component list";
+		if (!componentSubType) {
+			subTypeSource = "control plane";
+			try {
+				componentSubType = await fetchComponentSubType({
+					orgHandler: org.handle,
+					projectId: project.id,
+					componentId: workspaceCompId,
+				});
+			} catch (err) {
+				ext.logError(`Failed to resolve the sub-type of component ${workspaceCompId}`, err as Error);
+			}
+		}
+		ext.log(
+			`Source component ${workspaceCompId} has sub-type '${componentSubType ?? "<unknown>"}' (from ${subTypeSource}), ` +
+				`isPrebuilt=${component?.metadata?.isPrebuilt}, foundInList=${!!component}`,
+		);
+
+		if (isMcpProxyFromExistingApi(componentSubType)) {
 			// if its an MCP proxy over an existing API, attach the sources to its existing track instead of
 			// creating a new component. This converts the proxy into an MCP server built from source.
 			return await handleMcpProxyRepositoryAttach(workspaceCompId, component, org, project, createParams[0], workspaceFsPath, gitRoot!);
@@ -511,7 +545,7 @@ async function handlePrebuiltComponentUpdate(
  */
 async function handleMcpProxyRepositoryAttach(
 	workspaceCompId: string,
-	component: ComponentKind,
+	component: ComponentKind | undefined,
 	org: Organization,
 	project: Project,
 	createParam: WICloudSubmitComponentsReq['createParams'][number],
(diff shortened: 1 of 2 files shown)


wi/wi-extension/src/cloud/cmds/create-component-cmd.ts
```diff
@@ -535,6 +562,12 @@ async function handleMcpProxyRepositoryAttach(
 		return result;
 	}
 
+	const repositorySubPath = relativePath(gitRoot, createParam.componentDir);
+	ext.log(
+		`Attaching git repository to MCP proxy component '${componentName}' (${workspaceCompId}) instead of creating a new component: ` +
+			`repo=${createParam.repoUrl}, branch=${createParam.branch}, subPath='${repositorySubPath}', org=${org.handle}, project=${project.handler}`,
+	);
```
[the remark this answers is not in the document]
pcnfernando: Repo url would not include the credentials

hasathcharu (approved):

wi/wi-extension/src/cloud/cmds/create-component-cmd.ts
```diff
@@ -260,14 +260,41 @@ export const submitCreateComponentHandler = async ({ createParams, org, project,
...
+			} catch (err) {
+				ext.logError(`Failed to resolve the sub-type of component ${workspaceCompId}`, err as Error);
+			}
+		}
+		ext.log(
+			`Source component ${workspaceCompId} has sub-type '${componentSubType ?? "<unknown>"}' (from ${subTypeSource}), ` +
+				`isPrebuilt=${component?.metadata?.isPrebuilt}, foundInList=${!!component}`,
+		);
+
+		if (isMcpProxyFromExistingApi(componentSubType) && component) {
```
NipunaRanasinghe: The `&& component` added in ff49c46 cancels out the fallback this PR adds.

The reason for going to the control plane is that the cached list may not have the component (or its sub-type). But if the control plane says this *is* an unconverted MCP proxy and the component isn't in the cached list, we now fall through to the create flow and make a new component next to the proxy. The user is left with a duplicate and a proxy that was never converted, which is hard to undo from the editor.

So either the component is always in the list (then the CP query is only needed for the empty `subType` case, and `&& component` is dead code worth a comment), or it isn't (then this is exactly the case the PR is meant to fix). What was ff49c46 working around?

If `handleMcpProxyRepositoryAttach` only needs the `ComponentKind` for `result.created` and the "view in console" link, could the CP query select `name`/`handler` too and build what it needs, instead of falling back to creating a component?

wi/wi-extension/src/cloud/cmds/create-component-cmd.ts
```diff
@@ -260,14 +260,41 @@ export const submitCreateComponentHandler = async ({ createParams, org, project,
...
+		if (!componentSubType) {
+			subTypeSource = "control plane";
+			try {
+				componentSubType = await fetchComponentSubType({
+					orgHandler: org.handle,
+					projectId: project.id,
+					componentId: workspaceCompId,
+				});
+			} catch (err) {
+				ext.logError(`Failed to resolve the sub-type of component ${workspaceCompId}`, err as Error);
```
NipunaRanasinghe, line 309: A network or control plane error is logged and swallowed here, so the code carries on and creates a new component. A failed lookup for an id that came from the MCP proxy flow isn't the same as "not a proxy", and the cost of getting it wrong is a duplicate component in the cloud.

Would prefer surfacing the failure to the user (or retrying) rather than silently taking the create path.

wi/wi-extension/src/cloud/cmds/create-component-cmd.ts
```diff
@@ -260,14 +260,41 @@ export const submitCreateComponentHandler = async ({ createParams, org, project,
 	const dotGit = await newGit?.getRepositoryDotGit(workspaceFsPath);
 	const repo = newGit.open(gitRoot, dotGit);
 
-	const workspaceCompId: string | null | undefined = ext.context.workspaceState.get("SOURCE_COMPONENT_ID");
+	const stateCompId: string | null | undefined = ext.context.workspaceState.get("SOURCE_COMPONENT_ID");
+	const envCompId: string | undefined = process.env.SOURCE_COMPONENT_ID;
```
NipunaRanasinghe: Reading the env var is the right fix, since nothing copies it into workspace state (`context-store.ts:339` only uses it as a filter). But it's never cleared: after a successful attach we only null the workspace state, so `stateCompId || envCompId` picks the same id up again on the next deploy in the same session.

The proxy branch won't re-fire (sub-type is `...WithSource` by then), but every later create in that session pays an extra full-project CP call, and the stale id still reaches the `isPrebuilt` branch below before any sub-type check. Worth clearing it, or tracking "already consumed" in memory.

wi/wi-extension/src/cloud/graphql/cp-graphql-client.ts
```diff
@@ -130,6 +129,34 @@ export async function executeCpGraphql<T>(query: string, operationName: string):
...
+ * it). So `ComponentKind.spec.subType` is always an empty string on this side and is useless for
+ * detection. Returns undefined when the component is not found.
+ */
+export async function fetchComponentSubType(params: {
+	orgHandler: string;
+	projectId: string;
+	componentId: string;
+}): Promise<string | undefined> {
+	const query = `query {
+  components(orgHandler: ${gqlString(params.orgHandler)}, projectId: ${gqlString(params.projectId)}, options: { filter: { withSystemComponents: true }}) {
```
NipunaRanasinghe: Two things about this call:

1. It fetches every component in the project and filters by id on the client. Fine if the schema has no id filter, but the doc comment explains why `getItem` can't be used, not why a filtered `components` query can't.
2. The caller runs it before `checkComponentLimitReached` and outside any `window.withProgress`, with a 60s timeout. A slow control plane stalls the deploy for up to a minute with no feedback in the UI. The attach mutation is wrapped in a progress notification; this one should be too, or use a shorter timeout.

wi/wi-extension/src/cloud/cmds/create-component-cmd.ts
```diff
@@ -260,14 +260,41 @@ export const submitCreateComponentHandler = async ({ createParams, org, project,
...
+	const workspaceCompId: string | null | undefined = stateCompId || envCompId;
+
```
NipunaRanasinghe: Nit: this line has a stray tab, and line 537 has a trailing space. Biome will strip both, so the format check may fail.

Also, ff49c46 removed the comments added in 781a920 that explained why both env and workspace state are read, and why the sub-type can come back empty. Those covered non-obvious behaviour and are worth keeping.

wi/wi-extension/src/cloud/graphql/cp-graphql-client.ts
pcnfernando: 1.  no id filter exists: ComponentsFilterOptions only offers type, types, limit, offset and withSystemComponents. So a filtered components query can't narrow by id, and fetching the project and filtering client-side was the only option for that query

wi/wi-extension/src/cloud/cmds/create-component-cmd.ts
pcnfernando: Addressed in 6572b9e5fb6e012a4424a85d67413aa88eb369af

wi/wi-extension/src/cloud/cmds/create-component-cmd.ts
pcnfernando: Addressed in https://github.com/wso2/product-integrator/commit/6572b9e5fb6e012a4424a85d67413aa88eb369af

wi/wi-extension/src/cloud/cmds/create-component-cmd.ts
pcnfernando: Addressed in https://github.com/wso2/product-integrator/commit/6572b9e5fb6e012a4424a85d67413aa88eb369af

wi/wi-extension/src/cloud/graphql/cp-graphql-client.ts
pcnfernando: 2. Addressed in 6572b9e

wi/wi-extension/src/cloud/cmds/create-component-cmd.ts
pcnfernando: Addressed in https://github.com/wso2/product-integrator/commit/6572b9e5fb6e012a4424a85d67413aa88eb369af

NipunaRanasinghe (approved):
Approving.

One thing worth confirming (not blocking): the lookup now runs whenever a source component id resolves, and in the cloud editor the env var is set for most sessions, not just MCP proxy ones. So a transient control plane error will now block a plain component deploy that previously would have gone through, and every such deploy shows the "Checking integration type..." notification. That is the trade-off I asked for and it is probably the right one, but it does apply well beyond the proxy flow — was that deliberate? If it feels too broad, one option is to only hard-fail when there is other evidence this is a proxy flow, and keep the old fall-through otherwise.

Two minor points:

- The synthesized `ComponentKind` is a partial object cast through `unknown`. It works because the only fields read downstream are `metadata.handler` (console link) and `metadata.name` (log text), and it never enters the data cache, so the empty `spec.source` is never dereferenced. A short comment listing the fields it has to carry would stop a future consumer from silently reading `{}`.
- No tests, and the PR description is still the unfilled template.

pcnfernando:
> Approving.
> 
> 
> 
> One thing worth confirming (not blocking): the lookup now runs whenever a source component id resolves, and in the cloud editor the env var is set for most sessions, not just MCP proxy ones. So a transient control plane error will now block a plain component deploy that previously would have gone through, and every such deploy shows the "Checking integration type..." notification. That is the trade-off I asked for and it is probably the right one, but it does apply well beyond the proxy flow — was that deliberate? If it feels too broad, one option is to only hard-fail when there is other evidence this is a proxy flow, and keep the old fall-through otherwise.

We only have this env populated for two types; prebuilt integrations and mcp proxies that are not yet bound to a user repo. Rest returns before reaching this level. Hence; it's deliberate.

PR #2327, opened 2026-09-04
fix(ci): derive the signing public key instead of configuring it twice

gigara:
## Purpose

Two changes to how the update-signing keys are configured and checked. Both concern the same failure: CI signing with one key while released clients pin another.

### 1. Derive the signing public key instead of configuring it twice

`COSIGN_PUBLIC_KEY` was a second secret holding the public half of `COSIGN_PRIVATE_KEY`. It can only agree or disagree with the private key, and when it disagreed the run still passed — every signature verified against `COSIGN_PUBLIC_KEY`, which is exactly the key that was wrong. The check could not see the mismatch that matters.

Both workflows now derive it:

```yaml
cosign public-key --key env://COSIGN_PRIVATE_KEY > signing.pub
grep -q 'BEGIN PUBLIC KEY' signing.pub
```

`WSO2_UPDATE_PUBLIC_KEY` is compared against that derived key, so the comparison proves the pinned key matches the key that actually signs. `COSIGN_PUBLIC_KEY` is dropped from both workflows and can be deleted from the repository secrets once this merges.

### 2. Check the publish path before the builds run, not after

`WSO2_UPDATE_PUBLIC_KEY` was optional, and its absence was silent in the worst direction: the comparison was skipped and `WSO2_UPDATE_REQUIRE_ARTIFACT_SIGNATURE` computed to `false`, so a green release shipped clients that pin no key and accept unverified artifacts, then published a source document for them.

`resolve-versions` now gates the run. It is the right place because `build-macos` and `build-windows` bake `WSO2_UPDATE_URL` and `WSO2_UPDATE_PUBLIC_KEY` into `product.json` via `update-product.sh`. A check in `publish-update-source` runs after those installers exist; all it can do is decline to publish a document for them. Two steps, both gated on the same conditions as `publish-update-source`'s own `if:` (`publish_update_source && build_packed_installers && publish_tag != ''`), so a run that would not publish is never failed for missing publish secrets:

- **Check the publish path is configured** — `AWS_UPDATE_S3_BUCKET`, `WSO2_UPDATE_ARTIFACTS_URL`, `WSO2_UPDATE_URL`, `WSO2_UPDATE_PUBLIC_KEY`, `COSIGN_PRIVATE_KEY` are present.
- **Verify the pinned client key matches the signing key** — derives from `COSIGN_PRIVATE_KEY` and compares in DER. Presence alone would leave the wrong-pair case to the late check; this also exercises `COSIGN_PASSWORD`, which otherwise burns a full build before failing at the derive step.

The comparison in `publish-update-source` is now redundant and removed. That job still derives `signing.pub` for signature verification.

`publish-components.yml` is a single job with no earlier phase, so its own preflight gained `WSO2_UPDATE_PUBLIC_KEY` and its comparison is likewise unconditional (still skipped on `dry_run`).

## Effect on existing builds

`dev-build.yml` and `daily-build.yml` never pass `publish_update_source`, so it defaults to `false` and both new steps are skipped — those builds still run with no update secrets at all. Only `release.yml`, which defaults it to `true`, reaches them.

## Verification

Two throwaway cosign keypairs, against the real step scripts:

| Case | Result |
|---|---|
| Pinned key is the public half of the signing key | passes |
| Pinned key is from the **other** pair | refused — the case the old public-vs-public check structurally could not see |
| Pinned key empty | refused by the `[ -s pinned.pem ]` guard, rather than dying at `openssl` with a parse error |
| Pinned key not valid base64 | refused |

The derived PEM is DER-identical to the `cosign.pub` that `cosign generate-key-pair` writes, so no re-keying is needed. Both workflows parse under `YAML.safe_load`, and the preflight exits 1 listing exactly the missing secret names and 0 when complete.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved signing verification for published updates and components by deriving the verification key from the configured private key.
  * Added early validation for required publishing configuration and ensured the pinned update key matches the signing key.
  * Applied consistent signature verification to source documents and mirrored artifacts.
  * Removed the need for a separately configured public-key secret.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1600,7 +1600,7 @@ jobs:
     env:
       # secrets can't be referenced from a step `if:`, so their presence is hoisted here — the same
       # reason HAS_AWS exists on the build jobs.
-      HAS_PINNED_KEY: ${{ secrets.WSO2_UPDATE_PUBLIC_KEY != '' && secrets.COSIGN_PUBLIC_KEY != '' }}
+      HAS_PINNED_KEY: ${{ secrets.WSO2_UPDATE_PUBLIC_KEY != '' }}
       # New builds publish to the channel chosen at dispatch (default insider); the
       # promote-channel workflow later copies a source document down the ladder.
       CHANNEL: ${{ inputs.channel || 'insider' }}
@@ -1691,33 +1691,47 @@ jobs:
         if: ${{ steps.targets.outputs.targets != '' }}
         uses: sigstore/cosign-installer@v3
 
+      # The public half is DERIVED from the signing key rather than configured separately. A second
+      # public-key secret can only agree or disagree with this one, and when it disagreed the
+      # signatures still verified in-run -- so the mismatch that matters (pinned key vs the key that
+      # actually signs) was the one the old check could not see.
+      - name: Derive the signing public key
+        if: ${{ steps.targets.outputs.targets != '' }}
+        run: |
+          set -euo pipefail
+          cosign public-key --key env://COSIGN_PRIVATE_KEY > signing.pub
+          grep -q 'BEGIN PUBLIC KEY' signing.pub
+          echo "derived the public half of COSIGN_PRIVATE_KEY"
+        env:
+          COSIGN_PRIVATE_KEY: ${{ secrets.COSIGN_PRIVATE_KEY }}
+          COSIGN_PASSWORD: ${{ secrets.COSIGN_PASSWORD }}
+
       # The client verifies artifacts against WSO2_UPDATE_PUBLIC_KEY, baked into product.json at
       # build time; CI signs with COSIGN_PRIVATE_KEY. Nothing downstream compares the two, so if they
       # are not the same key pair every client rejects every artifact — and the signing steps below
-      # still pass, because they verify against COSIGN_PUBLIC_KEY. That failure would only surface on
+      # still pass, because they verify against the signing key itself. That failure would only surface on
       # a user's machine, after the release. One base64 decode catches it here instead.
       - name: Check the pinned key matches the signing key
         if: ${{ steps.targets.outputs.targets != '' && env.HAS_PINNED_KEY == 'true' }}
         run: |
           set -euo pipefail
           printf '%s' "$WSO2_UPDATE_PUBLIC_KEY" | base64 -d > pinned.pem 2>/dev/null || {
-            echo "::error::WSO2_UPDATE_PUBLIC_KEY is not valid base64. It must be 'base64 < cosign.pub'." >&2
+            echo "::error::WSO2_UPDATE_PUBLIC_KEY is not valid base64. It must be 'base64 -w0 < cosign.pub'." >&2
             exit 1
           }
-          printf '%s' "$COSIGN_PUBLIC_KEY" > signing.pem
-          # Compare the DER of each, so PEM line wrapping or a trailing newline cannot fail a
-          # correctly configured pair.
+          # Compared against the key derived from COSIGN_PRIVATE_KEY, so this proves the pinned key
+          # matches the key that actually signs. DER, so PEM wrapping or a trailing newline cannot
+          # fail a correctly configured pair.
           openssl pkey -pubin -in pinned.pem  -outform DER -out pinned.der
-          openssl pkey -pubin -in signing.pem -outform DER -out signing.der
+          openssl pkey -pubin -in signing.pub -outform DER -out signing.der
           if ! cmp -s pinned.der signing.der; then
-            echo "::error::WSO2_UPDATE_PUBLIC_KEY does not match COSIGN_PUBLIC_KEY. Clients built here would reject every artifact this run signs." >&2
+            echo "::error::WSO2_UPDATE_PUBLIC_KEY is not the public half of COSIGN_PRIVATE_KEY. Clients built here would reject every artifact this run signs." >&2
             exit 1
           fi
           echo "Pinned client key matches the signing key."
-          rm -f pinned.pem signing.pem pinned.der signing.der
+          rm -f pinned.pem pinned.der signing.der
         env:
           WSO2_UPDATE_PUBLIC_KEY: ${{ secrets.WSO2_UPDATE_PUBLIC_KEY }}
-          COSIGN_PUBLIC_KEY: ${{ secrets.COSIGN_PUBLIC_KEY }}
 
       - name: Sign source document
         if: ${{ steps.targets.outputs.targets != '' }}
@@ -1729,11 +1743,7 @@ jobs:
       - name: Verify source signature
         if: ${{ steps.targets.outputs.targets != '' }}
         run: |
-          printf '%s' "$COSIGN_PUBLIC_KEY" > cosign.pub
-          cosign verify-blob --key cosign.pub --signature source.json.sig source.json
-          rm -f cosign.pub
-        env:
-          COSIGN_PUBLIC_KEY: ${{ secrets.COSIGN_PUBLIC_KEY }}
+          cosign verify-blob --key signing.pub --signature source.json.sig source.json
 
       - name: Upload source artifact
         if: ${{ steps.targets.outputs.targets != '' }}
@@ -1764,7 +1774,7 @@ jobs:
           fi
           # Public key as a file, matching the source verify step: cosign's env:// keyref is used
           # for the private key, and the proven pattern here for verification is a key file.
-          printf '%s' "$COSIGN_PUBLIC_KEY" > cosign.pub
+
           # Sign the STATEMENTS the generator wrote, not the artifacts themselves. A signature over
           # artifact bytes proves only "WSO2 produced this file", which would leave the document free
           # to offer an old signed artifact under a new version label; the statement binds
@@ -1774,10 +1784,9 @@ jobs:
             cosign sign-blob --key env://COSIGN_PRIVATE_KEY "${file}" --output-signature "${file}.sig" --yes
             # Verify immediately: a signature that cannot be checked here would ship as one the
             # client rejects, turning a signing misconfiguration into a failed update for users.
-            cosign verify-blob --key cosign.pub --signature "${file}.sig" "${file}"
+            cosign verify-blob --key signing.pub --signature "${file}.sig" "${file}"
             echo "signed $(basename "${file}")"
           done
-          rm -f cosign.pub
           # Every mirrored artifact must have one: an unsigned artifact would be served with a
           # statement URL that 404s, which the client treats as a hard failure.
           artifacts=$(find artifacts-mirror -type f ! -name '*.statement.json' ! -name '*.sig' | wc -l | tr -d ' ')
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -1600,7 +1600,7 @@ jobs:
     env:
       # secrets can't be referenced from a step `if:`, so their presence is hoisted here — the same
       # reason HAS_AWS exists on the build jobs.
-      HAS_PINNED_KEY: ${{ secrets.WSO2_UPDATE_PUBLIC_KEY != '' && secrets.COSIGN_PUBLIC_KEY != '' }}
+      HAS_PINNED_KEY: ${{ secrets.WSO2_UPDATE_PUBLIC_KEY != '' }}
```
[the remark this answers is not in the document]
gigara: Valid, and fixed in 68e50fb — with one change of placement.

Failing in `publish-update-source` would be too late to prevent the harm. `build-macos` and `build-windows` bake `WSO2_UPDATE_URL` and `WSO2_UPDATE_PUBLIC_KEY` into `product.json` via `update-product.sh` long before that job runs, and `WSO2_UPDATE_REQUIRE_ARTIFACT_SIGNATURE` is computed from the same secret. By the time the publish job could refuse, the run has already produced installers that pin no key and skip signature verification; all a late check can do is decline to publish a document for them.

So the preflight went into `resolve-versions`, which runs before `compile` and the build jobs, gated on `publish_update_source == true`. It covers `AWS_UPDATE_S3_BUCKET`, `WSO2_UPDATE_ARTIFACTS_URL`, `WSO2_UPDATE_URL`, `WSO2_UPDATE_PUBLIC_KEY` and `COSIGN_PRIVATE_KEY`, and fails in seconds instead of after a full build.

With presence guaranteed, `HAS_PINNED_KEY` is gone from both workflows and the DER comparison is unconditional (`publish-components.yml` still skips it on `dry_run`). That workflow's own preflight now requires `WSO2_UPDATE_PUBLIC_KEY` too, as suggested.

.github/workflows/build.yml
```diff
@@ -1764,7 +1799,7 @@ jobs:
           fi
           # Public key as a file, matching the source verify step: cosign's env:// keyref is used
           # for the private key, and the proven pattern here for verification is a key file.
-          printf '%s' "$COSIGN_PUBLIC_KEY" > cosign.pub
+
```
NipunaRanasinghe: Stale comment plus a stray blank line. The `printf '%s' "$COSIGN_PUBLIC_KEY" > cosign.pub` these two lines describe was deleted, so the comment now explains code that isn't here. Drop all three lines.

.github/workflows/build.yml
```diff
@@ -183,6 +183,34 @@ jobs:
...
       ballerina_jre_version: ${{ steps.read-versions.outputs.ballerina_jre_version }}
     steps:
+      # Checked here rather than in Publish Update Source, because build-macos and build-windows
+      # bake WSO2_UPDATE_URL and WSO2_UPDATE_PUBLIC_KEY into product.json long before that job
+      # runs. A late check cannot stop a run from shipping installers that pin no key and skip
+      # signature verification; this one can, and it costs seconds instead of a full build.
+      # Conditions mirror the Publish Update Source job's own `if:`, so a run that would not
+      # publish anyway is never failed for missing publish secrets.
+      - name: Check the publish path is configured
+        if: ${{ inputs.publish_update_source == true && inputs.build_packed_installers == true && inputs.publish_tag != '' }}
```
NipunaRanasinghe, line 193: This step is new in this PR (confirmed absent from the base commit) and isn't mentioned anywhere in the PR description, which covers only the key derivation. It's ~28 of the 82 added lines and has the widest blast radius of anything here: it can now hard-fail a release at the very first job.

Please either document it in the description or split it into its own PR.

On blast radius, for the record: `publish_update_source` defaults to `false` in this workflow, and daily-build.yml, dev-build.yml and pr-ci.yml don't set it, so only release.yml (which defaults it to `true`) reaches this step. The `if:` conditions do mirror `publish-update-source`'s input conditions exactly, as the comment claims.

.github/workflows/build.yml
```diff
@@ -183,6 +183,34 @@ jobs:
...
+        run: |
+          set -euo pipefail
+          missing=()
+          [ -n "${AWS_UPDATE_S3_BUCKET}" ] || missing+=("AWS_UPDATE_S3_BUCKET")
+          [ -n "${WSO2_UPDATE_ARTIFACTS_URL}" ] || missing+=("WSO2_UPDATE_ARTIFACTS_URL (artifacts must be mirrored; clients refuse upstream hosts)")
+          [ -n "${WSO2_UPDATE_URL}" ] || missing+=("WSO2_UPDATE_URL (clients built here would have no update endpoint)")
+          [ -n "${WSO2_UPDATE_PUBLIC_KEY}" ] || missing+=("WSO2_UPDATE_PUBLIC_KEY (clients built here would pin no key and accept unverified artifacts)")
```
NipunaRanasinghe, line 202: This only checks that the secret is non-empty, so the failure this PR is actually about — pinned key present but from the wrong pair — is still caught only in `publish-update-source`, after the full macOS and Windows builds have run.

That's the case the step's own comment argues against: `build-macos` (L472-473) and `build-windows` (L1015-1016) bake `WSO2_UPDATE_PUBLIC_KEY` into product.json, so a mismatched key ships into installers well before the late check fires.

Adding `sigstore/cosign-installer` here and running the derive-and-compare in this job costs a few seconds and moves the whole check ahead of the build. The check in `publish-update-source` could then be dropped or kept as a cheap re-assert.

.github/workflows/build.yml
```diff
@@ -183,6 +183,34 @@ jobs:
...
+            printf '::error::missing required secret: %s\n' "${missing[@]}" >&2
+            exit 1
+          fi
+          echo "publish path configured for channel ${{ inputs.channel || 'insider' }}"
+        env:
+          AWS_UPDATE_S3_BUCKET: ${{ secrets.AWS_UPDATE_S3_BUCKET }}
+          WSO2_UPDATE_ARTIFACTS_URL: ${{ secrets.WSO2_UPDATE_ARTIFACTS_URL }}
+          WSO2_UPDATE_URL: ${{ secrets.WSO2_UPDATE_URL }}
+          WSO2_UPDATE_PUBLIC_KEY: ${{ secrets.WSO2_UPDATE_PUBLIC_KEY }}
+          COSIGN_PRIVATE_KEY: ${{ secrets.COSIGN_PRIVATE_KEY }}
```
NipunaRanasinghe, line 216: `COSIGN_PASSWORD` isn't in the required list, but the new "Derive the signing public key" step needs it. If the signing key is encrypted and that secret is missing, the run still burns a full build before dying at the derive step — the exact gap this check exists to close.

It can't be required unconditionally, since an unencrypted key legitimately has an empty password. But if the org's key is encrypted (cosign's `generate-key-pair` always encrypts), adding it here is worth it.

.github/workflows/build.yml
```diff
@@ -1691,33 +1716,47 @@ jobs:
...
+      # still pass, because they verify against the signing key itself. That failure would only surface on
       # a user's machine, after the release. One base64 decode catches it here instead.
       - name: Check the pinned key matches the signing key
-        if: ${{ steps.targets.outputs.targets != '' && env.HAS_PINNED_KEY == 'true' }}
+        if: ${{ steps.targets.outputs.targets != '' }}
         run: |
           set -euo pipefail
           printf '%s' "$WSO2_UPDATE_PUBLIC_KEY" | base64 -d > pinned.pem 2>/dev/null || {
-            echo "::error::WSO2_UPDATE_PUBLIC_KEY is not valid base64. It must be 'base64 < cosign.pub'." >&2
+            echo "::error::WSO2_UPDATE_PUBLIC_KEY is not valid base64. It must be 'base64 -w0 < cosign.pub'." >&2
```
NipunaRanasinghe: Nit: this error message is unreachable for the empty-secret case. `printf '%s' "" | base64 -d` exits 0 and writes an empty file, so an empty `WSO2_UPDATE_PUBLIC_KEY` skips this handler and dies on the `openssl pkey` line below with an opaque error instead.

Not a live bug — the new preflight catches emptiness first — but it makes the friendly message depend on that coupling. A `[ -s pinned.pem ] || { echo "::error::WSO2_UPDATE_PUBLIC_KEY is empty." >&2; exit 1; }` after the decode decouples it.

Same shape at publish-components.yml L332.

NipunaRanasinghe (dismissed):

.github/workflows/build.yml
gigara: Fixed — the two comment lines and the blank line are gone. They described the `printf '%s' "$COSIGN_PUBLIC_KEY" > cosign.pub` that this PR deletes.

.github/workflows/build.yml
gigara: Fair — the description covered only the derivation. I've rewritten it: the preflight is now section 2, with why it sits in `resolve-versions` rather than `publish-update-source`, the exact secrets it requires, and an explicit note that `dev-build.yml`/`daily-build.yml` never reach it.

Keeping it here rather than splitting, because it isn't independent of the derivation. Removing `COSIGN_PUBLIC_KEY` makes `WSO2_UPDATE_PUBLIC_KEY` the only thing left that can disagree with the signing key, so the pinned-key comparison stops being one of two overlapping checks and becomes the check. Landing that without also making it mandatory would leave a window where the sole remaining guard is optional.

Thanks for confirming the reachability and the `if:` conditions independently.

.github/workflows/build.yml
gigara: Agreed, and done. `resolve-versions` now installs cosign and runs the full derive-and-compare:

```yaml
cosign public-key --key env://COSIGN_PRIVATE_KEY > signing.pub
printf '%s' "$WSO2_UPDATE_PUBLIC_KEY" | base64 -d > pinned.pem || ...
openssl pkey -pubin -in pinned.pem  -outform DER -out pinned.der
openssl pkey -pubin -in signing.pub -outform DER -out signing.der
cmp -s pinned.der signing.der || fail
```

You're right that the non-emptiness check alone left the wrong-pair case exactly where the step's comment argues it shouldn't be — after L472-473 and L1015-1016 have written the key into product.json.

I dropped the late comparison rather than keeping it as a re-assert: the run cannot get past `resolve-versions` with a mismatched pair, so the second copy could only fire on a secret changed mid-run. `publish-update-source` still derives `signing.pub`, which it needs for `cosign verify-blob`.

Verified against two throwaway EC keypairs: correct pair passes, other pair refused, empty refused, non-base64 refused.

.github/workflows/build.yml
gigara: This resolves itself now that the derivation moved into `resolve-versions` — `cosign public-key --key env://COSIGN_PRIVATE_KEY` runs there with `COSIGN_PASSWORD` in its env, so a missing or wrong password fails in the first job rather than after a full build.

That's better than adding it to the required list, for the reason you note: it can't be required unconditionally, and exercising the secret proves more than asserting it is non-empty — a *wrong* password is caught too, which a presence check would miss. Called out in the step's comment.

.github/workflows/build.yml
gigara: Fixed in both places, with your guard:

```bash
[ -s pinned.pem ] || {
  echo "::error::WSO2_UPDATE_PUBLIC_KEY decoded to nothing." >&2
  exit 1
}
```

Right that it was unreachable — `printf '%s' "" | base64 -d` exits 0 and writes nothing. Worth decoupling even though the preflight catches emptiness first, since `publish-components.yml` has no earlier phase to rely on.

madushajg (dismissed):

NipunaRanasinghe (approved):

PR #2328, opened 2026-09-04
fix(ci): drop the retired installer bucket and restore the update-bucket login

gigara:
## Purpose

Two loose ends left by the merge of the update mechanism on top of upstream's removal of the old S3 installer uploads.

### 1. `AWS_S3_BUCKET` is dead config

`AWS_S3_BUCKET` was the **old installer bucket**. Upstream's `f7917ab` ("Remove uploading artifacts to aws s3 and aws tool calls in the code-build") removed the steps that wrote to it. What survived is `HAS_INSTALLER_BUCKET`, computed in the macOS and Windows job env blocks and read by **nothing**:

```
$ git grep -n HAS_INSTALLER_BUCKET
.github/workflows/build.yml:292:      HAS_INSTALLER_BUCKET: ${{ secrets.AWS_S3_BUCKET != '' }}
.github/workflows/build.yml:834:      HAS_INSTALLER_BUCKET: ${{ secrets.AWS_S3_BUCKET != '' }}
```

Installers now ship as GitHub release assets, and every remaining `aws s3` call in the repo targets `AWS_UPDATE_S3_BUCKET`. Removed both, plus the Windows job's `HAS_AWS` — that job no longer touches AWS at all.

### 2. The macOS Squirrel upload had no credentials — a real break

`f7917ab` also removed **both** `Login to AWS` steps. The update-mechanism merge restored only the one in `Publish Update Source`. The macOS *Publish Squirrel.Mac update payload* step (added while the mac job still had a login) was therefore running `aws s3 cp` with nothing configured:

- on a publishing release, `Unable to locate credentials` fails the macOS build job;
- the darwin manifest's `app.squirrel.url` — embedded by `generate-update-manifest.mjs` — would point at an object that was never uploaded, so macOS app updates would 404 at apply time.

This PR configures the credentials immediately before that upload (same region resolution as the publish job), rather than for the whole job, since those two steps are the only AWS consumers in it.

### 3. Tightened the gate

Both steps are now gated on `publish_update_source == true` and on the update bucket being configured, which replaces the in-script `-z` bucket check. Nothing writes to the update bucket unless the run is a release that publishes updates — `release.yml` defaults it to `true`, `dev-build.yml` and `daily-build.yml` don't pass it, so they get `false`.

## Secrets impact

`AWS_S3_BUCKET` can be **deleted** from the repository secrets after this merges. `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` must be the credentials for the **update** bucket — they need `s3:GetObject` and `s3:PutObject` on `manifests/*` and `artifacts/*` there.

## Verification

- `YAML.safe_load` parses the workflow; `build-macos.env` is `{HAS_AWS, HAS_UPDATE_BUCKET}` and `build-windows` no longer has an `env` key.
- `git grep AWS_S3_BUCKET` (excluding `AWS_UPDATE_S3_BUCKET`) returns nothing repo-wide.
- The only two AWS-touching steps in `build-macos` are the new login and the Squirrel upload.

## Summary by CodeRabbit

- **Build and Release**
  - macOS update payload publishing now follows the configured update-source setting alongside packed-installer builds.
  - Builds now report a failure when required publishing credentials are missing, instead of silently skipping the update upload.
  - Updated build configuration applies consistent handling of publishing-related settings across macOS and Windows packaging jobs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -285,11 +285,6 @@ jobs:
     name: Build (macOS)
     runs-on: ${{ matrix.runner }}
     needs: [resolve-versions, compile]
-    env:
-      # secrets can't be referenced from a step `if:`, so their presence is hoisted here. Both are
-      # needed to publish the Squirrel.Mac payload: credentials to write, and a bucket to write to.
-      HAS_AWS: ${{ secrets.AWS_ACCESS_KEY_ID != '' }}
-      HAS_UPDATE_BUCKET: ${{ secrets.AWS_UPDATE_S3_BUCKET != '' }}
     defaults:
       run:
         working-directory: lib/vscode
@@ -757,31 +752,41 @@ jobs:
             xcrun stapler staple "$ARTIFACT"
           done
 
-      # Only the two steps below touch AWS in this job, and only on a publishing release, so the
-      # credentials are configured immediately before the upload rather than for the whole job.
-      - name: Configure AWS credentials for the update bucket
-        if: ${{ inputs.publish_update_source == true && inputs.build_packed_installers == true && env.HAS_AWS == 'true' && env.HAS_UPDATE_BUCKET == 'true' }}
-        uses: aws-actions/configure-aws-credentials@v4
-        with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          aws-region: ${{ secrets.AWS_REGION || 'us-east-1' }}
-
+      # This is the only AWS consumer in the job, so the credentials go in its own `env:` rather
+      # than through configure-aws-credentials, which writes them to $GITHUB_ENV and would leave
+      # live S3 write credentials in scope for every step that follows.
+      #
+      # It fails rather than skips on a half-configured run: publish-update-source embeds
+      # app.squirrel.url into the darwin manifest regardless, so skipping here would ship a
+      # manifest pointing at an object nobody uploaded — a 404 for every macOS client at apply
+      # time, from a green build.
       - name: Publish Squirrel.Mac update payload to update bucket
-        if: ${{ inputs.publish_update_source == true && inputs.build_packed_installers == true && env.HAS_AWS == 'true' && env.HAS_UPDATE_BUCKET == 'true' }}
+        if: ${{ inputs.publish_update_source == true && inputs.build_packed_installers == true }}
         working-directory: .
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          AWS_DEFAULT_REGION: ${{ secrets.AWS_REGION || 'us-east-1' }}
+          AWS_UPDATE_S3_BUCKET: ${{ secrets.AWS_UPDATE_S3_BUCKET }}
         run: |
+          set -euo pipefail
+          missing=()
+          [ -n "${AWS_UPDATE_S3_BUCKET}" ] || missing+=("AWS_UPDATE_S3_BUCKET")
+          [ -n "${AWS_ACCESS_KEY_ID}" ] || missing+=("AWS_ACCESS_KEY_ID")
+          [ -n "${AWS_SECRET_ACCESS_KEY}" ] || missing+=("AWS_SECRET_ACCESS_KEY")
+          if [ ${#missing[@]} -gt 0 ]; then
+            printf '::error::missing required secret: %s\n' "${missing[@]}" >&2
+            exit 1
+          fi
           V="${{ needs.resolve-versions.outputs.integrator_version }}"
           ZIP="installers/mac/wso2-integrator-${V}-${{ matrix.arch }}-mac.zip"
-          # The step only runs when this build publishes updates, so a missing payload is the
-          # failure it looks like rather than a deployment that simply uploads nothing.
           if [ ! -f "$ZIP" ]; then
-            echo "Squirrel.Mac payload not found at $ZIP" >&2
+            echo "::error::Squirrel.Mac payload not found at $ZIP" >&2
             exit 1
           fi
           # Same artifacts layout as everything else; the darwin manifest's app.squirrel.url
           # (embedded by generate-update-manifest.mjs) points at this exact versioned CDN URL.
-          DEST="s3://${{ secrets.AWS_UPDATE_S3_BUCKET }}/artifacts/app/${V}/"
+          DEST="s3://${AWS_UPDATE_S3_BUCKET}/artifacts/app/${V}/"
           aws s3 cp "$ZIP" "${DEST}wso2-integrator-${V}-${{ matrix.arch }}-mac.zip" --content-type application/zip --cache-control no-cache
           echo "Published Squirrel.Mac payload for darwin/${{ matrix.arch }} (v${V})"
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -286,10 +286,10 @@ jobs:
     runs-on: ${{ matrix.runner }}
     needs: [resolve-versions, compile]
     env:
-      # secrets can't be referenced from a step `if:`. HAS_AWS says credentials exist;
-      # HAS_INSTALLER_BUCKET says a target bucket does. They are different secrets.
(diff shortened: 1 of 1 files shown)


gigara:
@coderabbitai review

.github/workflows/build.yml
```diff
@@ -756,18 +756,25 @@ jobs:
...
+
+      # Only the two steps below touch AWS in this job, and only on a publishing release, so the
+      # credentials are configured immediately before the upload rather than for the whole job.
+      - name: Configure AWS credentials for the update bucket
+        if: ${{ inputs.publish_update_source == true && inputs.build_packed_installers == true && env.HAS_AWS == 'true' && env.HAS_UPDATE_BUCKET == 'true' }}
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.AWS_REGION || 'us-east-1' }}
```
NipunaRanasinghe: "configured immediately before the upload rather than for the whole job" isn't what this does. `aws-actions/configure-aws-credentials` exports `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_DEFAULT_REGION` and `AWS_REGION` into the job environment, so they persist for every step after it — not just the next one.

Six steps follow the upload in this job: Upload macOS DMG artifacts (L788), cosign-installer (L798), Sign the dmg installer (L801), Verify signature (L807), Generate Hashes (L815), Upload Signatures and Hashes (L821). All of them run with live S3 write credentials in env, including `cosign-installer`, which fetches a third-party binary.

The placement is later than a job-level `env:` block, which is an improvement, but it isn't step-scoped. Two ways to get what the comment claims:

- move the Squirrel upload to the end of the job, so nothing runs after it; or
- drop the action and set `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` / `AWS_DEFAULT_REGION` in the upload step's own `env:` block — that genuinely scopes them to the one step, and it's a single `aws s3 cp` with no other AWS consumer to serve.

Either way the comment needs rewording.

.github/workflows/build.yml
```diff
@@ -756,18 +756,25 @@ jobs:
...
           V="${{ needs.resolve-versions.outputs.integrator_version }}"
           ZIP="installers/mac/wso2-integrator-${V}-${{ matrix.arch }}-mac.zip"
-          # Bucket first: a deployment that publishes no updates skips cleanly here, and only a
-          # deployment that WOULD upload treats a missing payload as the failure it is.
-          if [ -z "${{ secrets.AWS_UPDATE_S3_BUCKET }}" ]; then
-            echo "AWS_UPDATE_S3_BUCKET not set — skipping Squirrel.Mac payload upload."
-            exit 0
-          fi
+          # The step only runs when this build publishes updates, so a missing payload is the
+          # failure it looks like rather than a deployment that simply uploads nothing.
```
NipunaRanasinghe: The gate is broader than this comment says. It runs when the build publishes updates **and** `HAS_AWS` **and** `HAS_UPDATE_BUCKET` — so with `publish_update_source == true` but `AWS_UPDATE_S3_BUCKET` or `AWS_ACCESS_KEY_ID` unset, both steps skip, the macOS job goes green, and `publish-update-source` still generates a darwin manifest whose `app.squirrel.url` points at an object nobody uploaded.

That's the 404-at-apply-time failure this PR is fixing, reached by a different misconfiguration. Not a regression — the `-z "${AWS_UPDATE_S3_BUCKET}"` check you removed also `exit 0`'d — but the replacement inherits the same silent hole while the comment asserts it doesn't.

\#2327's new preflight in `resolve-versions` closes it by hard-failing on a missing `AWS_UPDATE_S3_BUCKET` before any build runs. Worth stating that dependency explicitly, or landing these two together — on its own this PR still lets a half-configured release ship a manifest pointing at nothing.

.github/workflows/build.yml
```diff
@@ -286,10 +286,10 @@ jobs:
     runs-on: ${{ matrix.runner }}
     needs: [resolve-versions, compile]
     env:
-      # secrets can't be referenced from a step `if:`. HAS_AWS says credentials exist;
-      # HAS_INSTALLER_BUCKET says a target bucket does. They are different secrets.
+      # secrets can't be referenced from a step `if:`, so their presence is hoisted here. Both are
+      # needed to publish the Squirrel.Mac payload: credentials to write, and a bucket to write to.
       HAS_AWS: ${{ secrets.AWS_ACCESS_KEY_ID != '' }}
-      HAS_INSTALLER_BUCKET: ${{ secrets.AWS_S3_BUCKET != '' }}
+      HAS_UPDATE_BUCKET: ${{ secrets.AWS_UPDATE_S3_BUCKET != '' }}
```
NipunaRanasinghe: Confirmed the removals are safe: no `HAS_INSTALLER_BUCKET` or `AWS_S3_BUCKET` reference survives anywhere in the repo, and after dropping the `env:` block from `build-windows` no step in that job reads `env.HAS_AWS`. The only remaining `env.HAS_AWS` readers are L763 and L771 here.

One leftover the diff doesn't reach: L1603-1604 in `publish-update-source` still says "the same reason `HAS_AWS` exists on the build job**s**". After this PR it exists on one job. Moot if #2327 merges first — that PR deletes both lines along with `HAS_PINNED_KEY`.

gigara:
@coderabbitai

madushajg (dismissed):

.github/workflows/build.yml
gigara: You're right, and the comment was the least of it — `configure-aws-credentials` writes to `$GITHUB_ENV`, so those six steps really were running with live S3 write credentials in scope.

Took your second option. The action is gone and the credentials live in the upload step's own `env:`:

```yaml
env:
  AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
  AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
  AWS_DEFAULT_REGION: ${{ secrets.AWS_REGION || 'us-east-1' }}
  AWS_UPDATE_S3_BUCKET: ${{ secrets.AWS_UPDATE_S3_BUCKET }}
```

One `aws s3 cp` with no other consumer in the job, so nothing needed the action's other behaviour. That's now step-scoped for real, and the comment says what it does. The bucket comes from `env:` too rather than `${{ secrets... }}` interpolated into the script.

.github/workflows/build.yml
gigara: Correct, and thanks for not letting the comment stand — the replacement did inherit the hole.

Rather than lean on #2327, the step now closes it itself. The `if:` no longer tests secret presence, and the step checks its own three secrets and fails:

```bash
missing=()
[ -n "${AWS_UPDATE_S3_BUCKET}" ] || missing+=("AWS_UPDATE_S3_BUCKET")
[ -n "${AWS_ACCESS_KEY_ID}" ] || missing+=("AWS_ACCESS_KEY_ID")
[ -n "${AWS_SECRET_ACCESS_KEY}" ] || missing+=("AWS_SECRET_ACCESS_KEY")
```

So a half-configured release cannot reach `publish-update-source` with the macOS job green. `HAS_AWS`/`HAS_UPDATE_BUCKET` then had no readers left and the macOS `env:` block went with them.

I did take your point about the earliest possible failure, in #2327: its `resolve-versions` preflight now requires `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` alongside `AWS_UPDATE_S3_BUCKET`, so the same misconfiguration fails in seconds instead of after the macOS build. The two are complementary now rather than one depending on the other — this PR is correct on its own.

Verified under bash 3.2 (the macOS runner's shell, where `set -u` and empty arrays can bite): all-missing lists three names, partial lists the rest, missing payload fails, fully configured passes.

.github/workflows/build.yml
gigara: Thanks for verifying the removals independently.

On L1603-1604: this PR now removes the last `HAS_AWS` too — the macOS `env:` block is gone along with the Windows one — so that comment is wrong regardless of merge order, not just stale in number.

I've left it untouched here on purpose. #2327 deletes those exact two lines along with `HAS_PINNED_KEY`, and editing them here would turn a clean delete into a merge conflict for whichever lands second. If #2327 slips, I'll drop them in a follow-up rather than leave a comment naming a variable that no longer exists.

NipunaRanasinghe (approved):

PR #2467, opened 2026-09-17
fix(update): heal partially seeded components and clean up user data on uninstall

gigara:
## Problem

On a fresh Windows install, a component could end up **permanently half-copied** under `~/.wso2-integrator/components/` (seen in the field with `ballerina-runtime`). Reinstalling the app didn't fix it; only manually deleting the directory and relaunching recovered the editor.

**Root cause:** first-run seeding runs in the `main.ts` bootstrap *before* the single-instance lock is claimed, and both racing instances (installer "launch now" + a shortcut double-click on the slow first boot) shared one staging directory `<version>.seeding`. The second instance's initial cleanup deleted the first instance's in-flight copy; the first instance then renamed a tree with holes into place, and the `existsSync(target)` idempotence check trusted it forever.

## Fix (in `patches/wso2ComponentUpdate.diff`)

- **Per-pid staging** (`<version>.seeding-<pid>`): instances can no longer touch each other's in-flight copy. The rename loser fails on "target exists" and cleans up only its own directory.
- **Completion marker** (`.wso2-seeded`, fsync'd, written last before the rename): the skip check is now *target exists AND marker present*. A marker-less directory — from the race, from power loss after the rename, or an already-broken seed on a machine in the field — is deleted and re-seeded on the next launch, so **affected installs self-heal without manual intervention**. If the bundled source is gone (editor-only install), the partial tree is kept since it cannot be rebuilt.
- **Stale staging sweep**: leftover `*.seeding*` directories whose owning process is dead (including the old pid-less format) are removed.
- **GC guard**: `garbageCollectComponentVersions` now skips a live-pid seeding directory instead of deleting a second instance's copy mid-flight, which would have re-opened the race from the other direction.
- 6 new unit tests in `wso2ComponentsManifest.test.ts` (fresh seed + marker, skip-when-complete, re-seed partial, keep partial when source is gone, stale sweep, pid liveness) — 9/9 pass. Also fixed the latent `-0` antisymmetry assertion in the existing `compareVersions` test in passing.

## Windows MSI: uninstall cleanup (`installers/windows`)

A real uninstall now removes `%USERPROFILE%\.wso2-integrator` (seeded runtimes, staged updates, installed extensions) via `util:RemoveFolderEx`, so users no longer have to find and delete the hidden folder by hand.

- **Upgrades keep user data**: the property is only set when `UPGRADINGPRODUCTCODE` is absent, and `RemoveFolderEx` no-ops on a blank property.
- **Opt-out**: `msiexec /x <msi> WSO2_KEEP_USERDATA=1` (the ARP uninstall runs without UI, so a checkbox is not possible in a plain MSI).
- macOS/Linux are intentionally untouched: there is no macOS uninstall hook, and deb/rpm scriptlets run as root where deleting per-user home data violates packaging policy. The harmful case there (a broken seed poisoning reinstalls) is fixed by the marker-based self-heal, which is platform-independent.

## Verification

- `npm run typecheck-client` (tsgo) passes.
- All 9 unit tests pass (`npm run test-node -- --run out/vs/platform/wso2Update/test/node/wso2ComponentsManifest.test.js`, Node 24).
- `quilt refresh` captured all lib/vscode changes into `wso2ComponentUpdate.diff`; WiX XML validated with xmllint.
- **Not verified here** (no Windows/WiX toolchain on this machine): an MSI build + install→uninstall (folder removed), install→upgrade (folder kept), and `WSO2_KEEP_USERDATA=1` (folder kept) smoke test on the next CI MSI.

## Summary by CodeRabbit

- **New Features**
  - Added automatic checking, downloading, verification, and staged installation of WSO2 Integrator runtime and extension updates.
  - Added update status indicators, download progress, commands, and configuration options.
  - Added runtime component versions to the About dialog.
  - Added safer startup fallback and recovery for unsuccessful runtime updates.
  - Bundled runtimes are now initialized and maintained automatically.

- **Bug Fixes**
  - Windows uninstall now removes WSO2 Integrator user data when appropriate, while upgrades preserve it.


patches/wso2ComponentUpdate.diff
```diff
@@ -2358,11 +2425,33 @@ Index: product-integrator/lib/vscode/src/vs/platform/wso2Update/node/wso2Compone
...
++				continue; // complete seed
++			}
++			if (!srcExists) {
++				// Editor-only install: nothing to re-seed from, and a pre-marker or partial
++				// tree beats no runtime at all. Do not delete what cannot be rebuilt.
++				log?.(`Seed ${id}@${version} lacks its completion marker but no bundled source remains; leaving it in place`);
++				continue;
++			}
++			try {
++				fs.rmSync(path.toNamespacedPath(target), { recursive: true, force: true });
```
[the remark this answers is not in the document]
gigara: Addressed in dbf5af3, with a lighter mechanism than a lock: the repair now **claims the marker-less target by atomically renaming it aside** to a private `<target>.repair-<pid>` name before deleting it. Rename is atomic, so exactly one process wins the repair, and the delete acts on a name no other instance touches — a process stalled between its marker check and its delete can no longer remove a target that a faster instance completed in between. Crashed repairs leave a `.repair-<pid>` dir that the existing stale sweep collects (dead-pid rule), and the GC skips live ones.

A cross-process lock file was considered and rejected: it reintroduces stale-lock recovery (crashed holder before the single-instance mutex exists), which is the same class of problem this protocol removes. Note also that even the unpatched worst case converged: the loser re-seeds a complete tree, and a copy failure leaves *no* target, which falls back to the bundled runtime that boot and re-seeds on the next launch — transient, unlike the permanent partial-commit this PR fixes.

patches/wso2ComponentUpdate.diff
```diff
@@ -2373,6 +2462,16 @@ Index: product-integrator/lib/vscode/src/vs/platform/wso2Update/node/wso2Compone
 +			fs.mkdirSync(path.dirname(target), { recursive: true });
 +			fs.cpSync(path.toNamespacedPath(srcDir), path.toNamespacedPath(staging), { recursive: true });
 +			ensureExecutablePermissions(staging);
++			// fsync the marker so a power-off cannot persist the rename while losing the
++			// marker; the copied payload itself is not fsync'd (too slow for hundreds of MB),
++			// so a torn payload after power loss still surfaces as marker-less → re-seeded.
++			const fd = fs.openSync(path.toNamespacedPath(path.join(staging, SEED_COMPLETE_MARKER)), 'w');
++			try {
++				fs.writeSync(fd, version);
++				fs.fsyncSync(fd);
```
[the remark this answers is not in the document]
gigara: Correct observation — the marker fsync makes the marker itself torn-proof but does not make the payload durable, and the original comment overstated the guarantee. Fixed the comment in dbf5af3 to document this as an **accepted residual risk** rather than closing it, because both proposed closures are worse than the window they remove:

- **Syncing the copied tree**: per-file fsync of a multi-hundred-MB, ~10k-file runtime tree adds tens of seconds to the first boot, synchronously in the bootstrap before any window exists.
- **Validating the payload against a durable manifest on later boots**: this would *false-positive and delete live data*, because component homes mutate at runtime — ICP writes its logs and its database into its own home (the rpm `%post` explicitly chmods `components/icp/bin/database/` and `logs/` for that), and the bootstrap creates a `dependencies` link inside the seeded Ballerina home. A stat/size/count mismatch is therefore expected on healthy installs, and acting on it would wipe an ICP database to "repair" a healthy seed.

The remaining exposure is a power cut in the OS write-back seconds after the rename — strictly narrower than the pre-PR behavior (where any interruption or race could commit a permanently trusted partial tree), and recoverable by the same manual delete that any power-loss corruption of user data needs.

installers/windows/WixPackage/IntegratorComponents.wxs
```diff
@@ -43,6 +43,27 @@
...
+         the hidden folder by hand. RemoveFolderEx only acts when its property has a value, so the
+         property is set ONLY on a real uninstall:
+         - during a major upgrade the old product is removed with UPGRADINGPRODUCTCODE set, and the
+           blank property makes the removal a no-op — user data survives app updates;
+         - `msiexec /x <msi> WSO2_KEEP_USERDATA=1` opts out of the cleanup (the ARP uninstall runs
+           without UI, so a checkbox cannot be offered there). -->
+    <Property Id="WSO2_KEEP_USERDATA" Secure="yes" />
+    <SetProperty Id="WSO2USERDATAFOLDER" Action="SetWso2UserDataFolder"
+      Value="[%USERPROFILE]\.wso2-integrator"
+      After="AppSearch"
```
NipunaRanasinghe: WiX schedules its own `WixRemoveFoldersEx` action after `AppSearch` as well. That action is what reads `WSO2USERDATAFOLDER` and builds the RemoveFile rows, so with both sitting on the same anchor the order comes from linker-assigned sequence numbers rather than from anything written here. If the custom action wins, the property is still empty when it runs, `RemoveFolderEx` does nothing, and `.wso2-integrator` survives the uninstall — the same outcome as the deliberate opt-out, with no log line to distinguish them.

`Before="AppSearch"` or `After="FindRelatedProducts"` removes the ambiguity. `UPGRADINGPRODUCTCODE` arrives on the command line when the old product is removed, so it is already set that early and the condition still works. Note that `Before="CostInitialize"` would be worse, not better, since CostInitialize comes after AppSearch.

Worth verifying on the next MSI build, since the PR notes no Windows toolchain was available.

patches/wso2ComponentUpdate.diff
```diff
@@ -2358,11 +2425,33 @@ Index: product-integrator/lib/vscode/src/vs/platform/wso2Update/node/wso2Compone
 +		if (!version) {
 +			continue;
 +		}
-+		const target = path.join(componentsHome, id, version);
-+		if (fs.existsSync(target) || !fs.existsSync(srcDir)) {
++		const componentDir = path.join(componentsHome, id);
++		const target = path.join(componentDir, version);
++		const marker = path.join(target, SEED_COMPLETE_MARKER);
```
NipunaRanasinghe, line 2507: The skip is now "target exists and marker present", and nothing in the field has a marker. So on the first launch after this ships, every existing install deletes and re-copies ballerina-runtime, icp and jre — synchronously in the main.ts bootstrap, before any window appears. Hundreds of megabytes, on Windows, through Defender.

This only gets skipped when the bundled component version changed in the same release, giving a fresh target path. On an editor-only patch release where runtime versions stay put, everyone pays it.

The same thing catches updater-installed trees: `installComponent` promotes to `components/<id>/<version>` without a marker, so if a downloaded version ever equals the bundled one, a verified download gets thrown away and replaced with a bundle copy.

One option is to stamp the marker in place instead of re-copying when a cheap check passes — the local manifest already knows which versions the app has booted healthy on. If you would rather just accept the cost, a log line on the re-seed decision at least makes it visible when support asks why the first launch took four minutes.

patches/wso2ComponentUpdate.diff
```diff
@@ -2358,11 +2425,33 @@ Index: product-integrator/lib/vscode/src/vs/platform/wso2Update/node/wso2Compone
...
++			}
++			if (!srcExists) {
++				// Editor-only install: nothing to re-seed from, and a pre-marker or partial
++				// tree beats no runtime at all. Do not delete what cannot be rebuilt.
++				log?.(`Seed ${id}@${version} lacks its completion marker but no bundled source remains; leaving it in place`);
++				continue;
++			}
++			try {
++				fs.rmSync(path.toNamespacedPath(target), { recursive: true, force: true });
++				log?.(`Removed incomplete seed ${id}@${version}; re-seeding`);
```
NipunaRanasinghe: This deletes the target and then copies. If the copy fails afterwards — disk full, an antivirus holding a handle — the catch removes staging and returns, and now the target does not exist at all. For a tree that was actually complete (see the comment above: every pre-marker install looks incomplete to this check), that turns a working install into no runtime. The old code could never reach that state.

Copying first and swapping afterwards avoids it: rename target aside, rename staging in, then delete the old one. That also closes the window where two instances both enter this branch and recursively remove the same tree at the same time.

patches/wso2ComponentUpdate.diff
```diff
@@ -2337,16 +2344,76 @@ Index: product-integrator/lib/vscode/src/vs/platform/wso2Update/node/wso2Compone
...
++	try {
++		process.kill(pid, 0);
++		return true;
++	} catch (err) {
++		// EPERM means the process exists but belongs to someone else — still alive.
++		return (err as NodeJS.ErrnoException).code === 'EPERM';
++	}
++}
++
++/** Removes leftover `*.seeding*` staging directories whose owning process is gone. */
```
NipunaRanasinghe: Two things about liveness here.

Pid reuse can make a staging directory permanent. A `<version>.seeding-1234` left behind by a power loss survives the reboot, and after the reboot pid 1234 belongs to some unrelated process. `isLiveSeedingDir` then returns true forever, so the directory is skipped by this sweep and skipped by `garbageCollectComponentVersions` as well, leaking a partial runtime copy that nothing will ever clean up. Windows recycles pids quickly, so this is not a far-fetched case. An age guard fixes it: sweep regardless of pid once the directory's mtime is more than a few hours old.

Separately, a pid-less `<version>.seeding` is always treated as stale, and during the upgrade window that name belongs to an old build that may still be running and mid-copy. The result is a failed seed retried on the next launch rather than corruption, since the old code renames only after copying — but this is the exact transition the change ships into, so better known than discovered.

patches/wso2ComponentUpdate.diff
```diff
@@ -2381,7 +2480,11 @@ Index: product-integrator/lib/vscode/src/vs/platform/wso2Update/node/wso2Compone
 +			try {
 +				fs.rmSync(path.toNamespacedPath(staging), { recursive: true, force: true });
 +			} catch { /* best effort */ }
-+			log?.(`Failed to seed ${id}@${version}: ${err}`);
++			if (fs.existsSync(marker)) {
++				log?.(`Another instance completed seeding ${id}@${version} first`);
```
NipunaRanasinghe: `err` is dropped on this branch. A real ENOSPC or EPERM that happens to coincide with another instance finishing gets logged as a harmless race, which is the one case where you would most want the underlying error in the log. Include `err` in both branches.

patches/wso2ComponentUpdate.diff
```diff
@@ -3572,6 +3679,86 @@ Index: product-integrator/lib/vscode/src/vs/platform/wso2Update/test/node/wso2Co
...
++		for (const dir of [stalePidless, staleDeadPid, liveOwnPid]) {
++			fs.mkdirSync(dir, { recursive: true });
++		}
++		seedBundledComponents(componentsHome, seeds());
++		strictEqual(fs.existsSync(stalePidless), false);
++		strictEqual(fs.existsSync(staleDeadPid), false);
++		strictEqual(fs.existsSync(liveOwnPid), true);
++	});
++
++	test('isLiveSeedingDir recognizes only live-pid staging names', () => {
```
NipunaRanasinghe: The two behaviours the fix actually rests on have no test: the rename loser cleaning up only its own staging when a second instance wins, and `garbageCollectComponentVersions` skipping a live seeding directory. `isLiveSeedingDir` in isolation is the easy half.

The GC skip in particular is cheap to cover — a fake manifest entry and a directory named with the current pid is enough to prove the guard is wired up. The rename loser is harder to test honestly, but pre-creating a complete target before the call at least exercises the failure path and the "another instance finished first" branch.

installers/windows/WixPackage/IntegratorComponents.wxs
gigara: Fixed in 614f1e0 — moved to `Before="AppSearch"`. Both gating properties are command-line-borne (`UPGRADINGPRODUCTCODE` from the upgrade's RemoveExistingProducts, `WSO2_KEEP_USERDATA` from the user), so the condition still evaluates correctly that early. Added a comment in the .wxs explaining why the anchor matters, and the PR checklist already flags the install→uninstall→upgrade smoke test for the next MSI build.

patches/wso2ComponentUpdate.diff
gigara: Fixed in 614f1e0 with the option you suggested: a marker-less target that still **contains** the bundled payload is adopted in place — marker stamped, no copy. Containment = every source entry exists in the target with the same kind, and no file that has content in the source is empty in the target; extra entries, sizes and contents are deliberately ignored because component homes mutate at runtime (ICP logs/database, the Ballerina `dependencies` link), so an equality check would reject healthy seeds. The one-time cost for existing installs drops from a multi-hundred-MB re-copy to a stat walk, and a verified updater download equal to the bundled version is adopted rather than discarded. Trees that fail containment log `Seed <id>@<v> is missing bundled payload; re-seeding` before the copy, so a slow first launch is attributable in the log. (I used the containment check rather than the manifest's healthy-boot record because seeded baselines are never `active` in the manifest — health tracking only covers activated overrides.)

patches/wso2ComponentUpdate.diff
gigara: Fixed in 614f1e0, reordered exactly as you describe: the replacement is fully staged (copy + perms + marker) before the existing tree is touched, then the old tree is claimed with an atomic rename, the staging renamed in, and the claim removed. If the swap-in fails, the claimed tree is renamed back, so a copy or swap failure leaves the machine no worse than it started. Combined with adoption (previous thread), the delete path now only ever runs on trees that provably lack bundled payload. Covered by a test: a failed re-seed (unwritable component dir, posix) leaves the pre-existing tree intact.

patches/wso2ComponentUpdate.diff
gigara: Fixed in 614f1e0: liveness is now `isActiveSeedingDir(parentDir, name)` = pid alive **and** directory mtime younger than 24h, used by both the sweep and the GC, so a pre-reboot leftover with a recycled pid is collected instead of leaking forever. Tested with an aged own-pid directory. The trade-off is noted in the code: the staging root's mtime is set when the copy starts, so a copy that genuinely spans a day (hibernation mid-first-boot) can lose its rename to a concurrent sweep and retries on the next launch.

On the pid-less `<version>.seeding` during the upgrade window: agreed it's benign-but-real — the old build renames only after its copy completes, so the worst case is that build's seed failing once and retrying (against a marker-stamped or adopted target, where it loses cleanly). Given it's a one-release transition window I've left it as a known case rather than adding grace logic for it.

patches/wso2ComponentUpdate.diff
gigara: Fixed in 614f1e0 — the race branch now logs `Another instance completed seeding <id>@<v> first (own attempt failed with: <err>)`, so a coinciding ENOSPC/EPERM is visible.

patches/wso2ComponentUpdate.diff
gigara: Partially addressed in 614f1e0. Added: adoption of a healthy marker-less tree, rejection of a truncated (zero-length) file, the keep-existing-tree-on-failed-re-seed path (posix — exercises the copy-first ordering), the aged/reused-pid sweep, and the full age+liveness matrix for `isActiveSeedingDir`.

Two gaps remain, deliberately: (1) the GC **wiring** — `Wso2ComponentUpdateService` imports `electron` directly, so it cannot be constructed from a node-layer unit test without violating the layer rules; the guard it calls is covered directly and the call site is one line. (2) An honest two-process rename-loser test — simulating a second instance winning *between* this process's staging and rename needs either process orchestration or fs fault injection; the closest single-process approximations are the adoption and failed-re-seed tests above. If you think either is worth the harness cost I'm happy to add an electron-main test suite for the service in a follow-up.

installers/windows/WixPackage/IntegratorComponents.wxs
NipunaRanasinghe: Confirmed — `Before="AppSearch"` removes the ambiguity, and the comment explains why the anchor matters, which is the part that keeps someone from "tidying" it back later. Nothing further here.

patches/wso2ComponentUpdate.diff
NipunaRanasinghe: Adoption looks right, and it does cover the actual field failure. Worth spelling out why: in the race you described, the second instance's `rmSync` removes files the first instance's `cpSync` has already written and will not revisit, so the committed tree has whole entries **missing**. Containment tests exactly that, so the trees this was built for are correctly rejected.

The gap is a file that exists but is truncated to a non-zero length — containment passes it and the marker gets stamped on a corrupt tree, permanently trusted again. I looked at tightening the file check to `dstSize < srcSize` and I don't think you should: user-editable files in a component home (config, properties) can legitimately shrink, so that predicate would delete user edits to fix a case that only arises from a power cut in the write-back window. `size === 0` is the right conservative line. Recording the reasoning so the next person doesn't have to redo it.

One operational note: the stat walk is one-time only if the stamp succeeds. If the component home is read-only or the write fails, the adoption path runs the full walk on every launch, synchronously in bootstrap. The log line is there, but it may be worth counting it as a failure rather than an informational message, since a permanently unstampable seed is a real state.

patches/wso2ComponentUpdate.diff
NipunaRanasinghe: The ordering is right, but the age guard and the repair claim interact badly.

`rename()` does not touch the mtime of the directory being renamed — it updates the parent directories. So `<target>.repair-<pid>`, created by renaming the existing seed aside, carries the **seed tree's** mtime, which on any real install is weeks or months old. `isActiveSeedingDir` therefore returns false for the claim from the moment it exists, even though the owning process is alive and mid-swap.

That matters on the restore path. If a concurrent instance runs `sweepStaleSeedingDirs` (or the GC) in the window between the claim and the swap, it deletes the claimed tree. If the staging rename then fails, the restore has nothing to rename back — no target at all, which is the exact outcome copy-first was added to prevent. The window is two syscalls wide so this is narrow, not urgent, but it is the same class of assumption as the one that produced the original bug.

Either `fs.utimesSync(claimed, now, now)` right after the claim rename, or skip the age check for `.repair-` names and let pid liveness alone decide — a repair claim only outlives its process on a crash, so it has no reason to need the age fallback. The tests won't catch this as written, since they create `.repair` directories with `mkdirSync` and get a fresh mtime for free; producing one by rename would reproduce it.

Separately, the comment above the swap claims more than the code does: "exactly one process wins, so a process stalled between its marker check and this point cannot remove a tree a faster instance completed." It can. An instance that started from a marker-less target, spent minutes copying, and arrives after a faster instance has already renamed a complete tree into place will find `target` existing, claim it, rename its own staging in, and delete the winner's tree. The content is identical so nothing breaks, but there is a brief window where `target` does not exist, and the invariant in the comment isn't the one being enforced. Re-checking `fs.existsSync(marker)` immediately before claiming would make the comment true and skip the pointless replacement.

patches/wso2ComponentUpdate.diff
NipunaRanasinghe: Age plus liveness is the right guard and the reused-pid test covers the case I raised. One knock-on effect: the same age check now also applies to `.repair-<pid>` claims, which inherit their mtime from the renamed tree and so read as stale immediately. Details on the swap thread.

patches/wso2ComponentUpdate.diff
gigara: Both points fixed in 9f72943 — thanks, the mtime interaction is a genuine bug I missed.

- **Claim mtime**: went with `utimesSync(claimed, now, now)` right after the claim rename rather than exempting `.repair-` from the age check — a pid-only exemption would re-open the reused-pid leak for claims stranded by a crash+reboot, which is exactly what the age guard exists for. Added a test that documents the rename-keeps-mtime trap by producing the claim via rename (stale immediately) and asserting the refresh restores protection.

- **Overclaiming comment / marker re-check**: the swap now re-checks `fs.existsSync(marker)` after the copy completes and discards its own staging when a faster instance finished — no pointless replacement, no transient no-target window for that case. The comment now states only what is enforced: one winner per claim, and a completion landing in the residual window is replaced with identical content.

patches/wso2ComponentUpdate.diff
gigara: Agreed on keeping `size === 0` as the line, and thanks for recording the shrinking-config rationale. The operational note is addressed in 9f72943: a failed adoption stamp now logs as an ERROR that names the recurring cost (`the containment walk will repeat on every launch until this succeeds`) instead of reading like an informational retry.

patches/wso2ComponentUpdate.diff
gigara: Fixed via the swap thread: the claim's mtime is refreshed with utimes immediately after the rename, so `.repair-` claims are protected while their owner lives and still age out after a crash+reboot with a recycled pid (9f72943).

patches/wso2ComponentUpdate.diff
NipunaRanasinghe: Both fixes check out, and the rename-keeps-mtime test is the right way to pin it — it asserts the trap before asserting the refresh, so it fails if someone removes the `utimes` later.

One marginal leftover in the same block: the `utimesSync` sits inside the claim's own `try`, whose `catch` sets `claimed = undefined`. If the refresh throws after the rename has already succeeded, the old tree is sitting at `.repair-<pid>` but the restore path no longer knows about it, so a subsequent staging-rename failure leaves no target — the case the claim exists to prevent. `utimesSync` on a directory you just produced by rename is about as reliable as a syscall gets, so this is close to theoretical; moving the refresh below the inner `catch`, or giving it its own swallowing `try`, closes it for free.

Nothing else outstanding from me on the seeding logic.

NipunaRanasinghe (approved):
LGTM

PR #2495, opened 2026-09-18
fix(installer): allow same-version MSI upgrades so 5.1.0-* builds replace each other

gigara:
## Problem (hit in the field, 2026-09-18)

`build.bat` folds the WiX `ProductVersion` from the release version **after stripping the pre-release suffix**, so `5.1.0-alpha4`, `5.1.0-09172` and `5.1.0-09181` all ship as MSI ProductVersion **`5.1.0`**. WiX `MajorUpgrade` only removes *lower-versioned* related products, so installing one `5.1.0-*` build over another registers a **second product** into the same per-user folder — and the in-app editor-update then stacks a **third**.

Observed sequence on a real machine (Event Log): alpha4 (9/16) → 09172 full (14:32) → in-app update applies `09181-update.msi` (14:40) → **three "product removed" events** when uninstalling (14:54–14:56). The stacked installs left a mixed Electron tree that died at launch with a Chromium **sandbox fatal** (`exit -36861` / `0xFFFF7003`, `crashpad not connected`, no window, no error), and the interleaved per-product `RemoveFolderEx` runs on uninstall corrupted the shared-component refcounts so **even full uninstall + reinstall of 09172 stayed broken**. Recovery required uninstalling every registered entry and manually deleting `%LOCALAPPDATA%\Programs\WSO2` before reinstalling.

Every 09172 user who accepts the in-app update to 09181 walks into this.

## Fix

`MajorUpgrade AllowSameVersionUpgrades="yes"` (+ `ICE61` suppression, which flags the intentionally self-inclusive upgrade range). Any build of the same release core now **replaces** the installed one — including the already-shipped broken builds, so this self-heals stacked machines on their next full-MSI install — and the eventual GA `5.1.0` still upgrades over every daily.

**Deliberately not fixed by folding the suffix into the version** (e.g. `5.1.9181`): that would compare *higher* than the GA's `5.1.0` and block the final-release upgrade. Trade-off accepted: MSI cannot order two dailies of the same core (installing an older daily over a newer one silently "succeeds"), which is fine within a pre-release train — the update server's semver comparison still decides what is offered.

Comments added in `Package.wxs` and `build.bat` tying the suffix-stripping and the flag together so neither is "fixed" independently.

## Verification

- XML validated; no Windows/WiX toolchain on this machine.
- Needs on the next MSI CI build: install `-09172` full → install a newer same-core MSI over it → **exactly one** product registered (Add/Remove shows one entry), app launches; then GA-style upgrade check (`5.1.0` over `5.1.0-*`).


samithkavishke (approved):
LGTM

PR #2501, opened 2026-09-19
Isolate the Agent Builder update feed and stop shipping update-only payloads as release assets

gigara:
> Includes #2500's commit (same base branch — GitHub cannot base a PR on a fork branch, so the diff is cumulative). **Review the last commit only**; merge in order #2500 → #2501.

## Problem

1. **Feed conflict risk**: the update client sends nothing product-identifying (`POST {channel, platform, arch, appVersion, components, bucket}`), and `WSO2_UPDATE_URL` is passed to `update-product.sh` unconditionally — an agent-builder build with the secret set polls the **integrator** feed and would be offered integrator installers. The S3 publish paths (`manifests/<channel>/`, `artifacts/app/<version>/`) also collide between flavors.
2. **Update-only payloads leak into public releases**: `wso2-integrator-<v>-update.msi` (+ sidecars) and both `*-mac.zip` Squirrel payloads are attached to public releases (see v5.1.0-alpha5, v5.1.0-09181). They are delivery payloads of the in-app updater, not user-installable artifacts (the mac zips are editor-only).
3. Latent bug found on the way: the Windows smoke test's `*.msi | Select -First 1` sorts `-update.msi` before `.msi` (`-` < `.`), so it has been installing the editor-only MSI — which ships no bundled runtimes.

## Change

- **Feed isolation, zero client change**: `update-product.sh` appends `/agent-builder` to `WSO2_UPDATE_URL` and `STOCK_UPDATE_URL` for the flavor. One secret serves both flavors; the client appends its API paths verbatim and its host allowlist ignores paths. The suffixed routes 404 until the update server grows matching product routes (component checks degrade to log-and-continue) — rollout is deliberately client-first.
- **Flavor-scoped publishing**: `manifests/agent-builder/<channel>/`, `artifacts/agent-builder/`, flavored `--artifacts-base`, and per-product `update-index-*` concurrency groups. `promote-channel.yml` gains a `product` input (the prefix rides inside the channel argument the promote script already composes paths from). `publish-components.yml` stays integrator-only for now — an AB release's own source document carries the full component set; a components-only lane for AB is deferred until an AB line outlives its bundled components.
- **De-leak**: `release-files/` is pruned of `*-update.msi*` and `*-mac.zip` before publishing, with negative assertions in Verify release assets as the belt. The generator gains `--app-installer-dir`: `publish-update-source` downloads the `Product-Integrator-Windows` workflow artifact and hashes/mirrors the `-update.msi` from disk (it previously fetched it from the release URL, which would now 404). Squirrel manifest URLs are CDN-only — publishing a squirrel target without `--artifacts-base` is a hard error instead of a silent fallback to a release asset that no longer exists (`--no-download` structure checks keep the fallback).
- **Smoke test** picks the full MSI explicitly.

## Verification

- Both flavors' `product.json` render verified: integrator URLs unchanged; AB gets `<base>/agent-builder` on both `wso2UpdateUrl` and `updateUrl`.
- Generator runs verified: `--app-installer-dir` resolves the local `-update.msi` to the CDN URL; components-only carry path still works; all YAML parses; `bash -n` clean.

Part 2 of 3. The update-server counterpart (product-segment routes) is on the `update-server-deploy` line.


dan-niles (approved):
LGTM

PR #2502, opened 2026-09-19
Publish Agent Builder releases into wso2/product-agent-builder

gigara:
> Includes #2500 and #2501's commits (same base branch, cumulative diff). **Review the last commit only**; merge in order #2500 → #2501 → #2502.

## Change

Agent Builder releases are created in the separate public repo `wso2/product-agent-builder`, with an independent version line. The source (and the release branch) stays here.

**release.yml**, when `product_flavor=agent-builder`:
- the release branch is `agent-builder-<version>` (hyphenated — `agent-builder/<version>` would D/F-conflict with the existing `agent-builder` branch), so the two products' version lines never collide in this repo;
- the exactly-once preflight checks the **branch here** and the **`v<version>` tag/release in `wso2/product-agent-builder`**;
- a new early preflight fails the run before an hour of building when `secrets.AGENT_BUILDER_RELEASE_TOKEN` (fine-grained PAT, `contents: write` on the release repo) is missing or cannot read it;
- `release_repository` is passed to build.yml — **only release.yml ever sets it**, so dev/nightly flavor runs can never publish cross-repo;
- `publish-marketplace` is gated off for agent-builder: the shared WI extension VSIX publishes to the marketplaces from integrator releases only.

**build.yml**:
- new `release_repository` input; `Create GitHub Release` passes `repository:`/`token:` to action-gh-release (`GITHUB_TOKEN` cannot write to another repo);
- cross-repo releases omit `target_commitish` (the built commit is not an object in the target repo; the tag lands on its default-branch HEAD) and record the source commit in the release body + job summary: `Built from wso2/product-integrator@<sha>`;
- the nightly tag-roll step is additionally gated on same-repo releases.

Release-workflow concurrency is keyed by flavor+version.

## Ops before first use

1. Create `AGENT_BUILDER_RELEASE_TOKEN` in this repo's secrets.
2. Dispatch `release.yml` with `ref=agent-builder`, `product_flavor=agent-builder`, `version=0.1.0(-m1)`.

Part 3 of 3.


.github/workflows/release.yml
```diff
@@ -137,6 +137,11 @@ jobs:
     runs-on: ubuntu-latest
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      # Agent Builder releases live in a separate public repo with their own independent
+      # version line; the source (and its release branch) stays in this repo. The release
+      # branch carries an agent-builder/ prefix so the two products' lines never collide.
+      RELEASE_REPO: ${{ inputs.product_flavor == 'agent-builder' && 'wso2/product-agent-builder' || github.repository }}
+      RELEASE_BRANCH: ${{ inputs.product_flavor == 'agent-builder' && format('agent-builder/{0}', inputs.version) || inputs.version }}
```
dan-niles: A branch named `agent-builder` already exists in this repo. `RELEASE_BRANCH` here evaluates to `agent-builder/<version>` for the agent-builder flavor, and this branch is pushed to `origin` (product-integrator itself, not product-agent-builder — the checkout above has no `repository:` override).

Git can't have both `refs/heads/agent-builder` and `refs/heads/agent-builder/<version>` at the same time (D/F ref conflict). Reproduced locally:
```
fatal: cannot lock ref 'refs/heads/agent-builder/0.1.0': 'refs/heads/agent-builder' exists; cannot create 'refs/heads/agent-builder/0.1.0'
```

So `git push origin "${RELEASE_BRANCH}"` (line 254) will fail the first time this is run per the PR's own "dispatch with `ref=agent-builder`" instructions — the branch-exists preflight above only checks for `agent-builder/<version>` itself, not this conflict with the parent `agent-builder` branch.

Needs either renaming/deleting the existing `agent-builder` branch, or a release-branch naming scheme that doesn't nest under it (e.g. `release/agent-builder-<version>`).

gigara: Good catch — confirmed: the slash form nests the ref under the existing `agent-builder` branch and git refuses the D/F conflict, so the first release push would have failed exactly as you reproduced. Fixed in edab643: the release branch is now hyphenated, `agent-builder-<version>` (no existing `agent-builder-*` or `release*` branches to collide with), with a comment on the env var explaining why a slash is off the table. PR body updated to match.

dan-niles (approved):
LGTM

